PR #15, opened 2025-01-22
Fetch Safe API to get transaction hash

iceanddust:


generatorpoint (approved):

inclinedmind (commented):
Should we add the max number of retries like coderabbit suggested?

codewiththeflowxyz (approved):
LGTM! @inclinedmind suggestion makes sense to me.

bitsforpeace (approved):

PR #20, opened 2025-01-23
Add Arbitrum E2E tests

iceanddust:
### What does this PR do?
- Refactor current Base tests to be L2 common tests and call this file for the Base and Arbitrum tests
- Add Arbitrum CI testnet config
- Update name of initial savings tests, to avoid regular expression from running other tests
- Update name of workflow files


inclinedmind (approved):
nice work!

PR #32, opened 2025-01-28
Improve CI workflows execution

iceanddust:
### What does this PR do?
- Implement fixtures for the webapp E2E tests so we have better control over the lifecycle of the test suite. Additions and modifications:
  - `waitForVnetsReady` is called once at the beginning of the worker
  - Account funding happens only once, immediately after the previous call
  - EVM Snapshot is taken after that
  - RPC mocking continues happening once before each test, but was moved to the fixture
  - After each test, the EVM is reverted to its initial state (post funding)
- Modify tests to use the new `test` function imported by the fixture and adjust tests to account for the above changes
- Update workflow versions used
- Improve caching of node modules and builds, this enables us now to only rerun failed steps in case of tests failing, instead of having to rerun the whole workflow. This already worked for the webapp E2E tests but not for the packages tests
- Increase CI timeout from 10 to 60 minutes. This makes the runs not fail if they run out of time (normally would take around 12 minutes if there are a couple of retries). And in case of failure, gives the workflow enough time to upload the Playwright reports and traces


.github/workflows/e2e.yml
```diff
@@ -41,12 +41,12 @@ jobs:
           - test: base-trade.spec
             vnet: true
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 60
```
generatorpoint, line 44: do we still want 60 minutes here? or maybe 20 would suffice

apps/webapp/src/test/e2e/fixtures.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export const test = playwrightTest.extend<TestFixture, WorkerFixture>({
+  // One-time setup fixture. This will run once at the beginning of the worker and provide the EVM snapshotIds to the tests or to other fixtures
+  snapshotIds: [
+    // eslint-disable-next-line no-empty-pattern
+    async ({}, use) => {
+      await waitForVnetsReady();
+
+      // Fund address before taking the snapshot, so we can return to a point where the address already has the funds
```
generatorpoint, line 34: nice, this should save some time during the test run

.github/workflows/e2e.yml
iceanddust: Yeah I think 20 would be more than enough. As tests become larger we may need more time, but we could make adjustments eventually

apps/webapp/playwright.config.ts
```diff
@@ -52,7 +52,7 @@ export default defineConfig({
     port: 3000,
     timeout: 120000,
     reuseExistingServer: !process.env.CI
-  },
-  globalSetup: './src/test/e2e/globalSetup',
-  globalTeardown: './src/test/e2e/globalTeardown'
+  }
+  // globalSetup: './src/test/e2e/globalSetup',
+  // globalTeardown: './src/test/e2e/globalTeardown'
```
generatorpoint, line 57: maybe we should remove these lines or add a comment why they're disabled if we want to keep them for potential future use

generatorpoint (approved):
lgtm!

apps/webapp/playwright.config.ts
iceanddust: I think I'll remove them. Those shouldn't be necessary anymore

PR #39, opened 2025-01-30
Advanced mode

bitsforpeace:
* Introduces `VITE_ADVANCED_CHAT_ENABLED` for enabling advanced mode (single endpoint fetch)
* Introduces `VITE_CHAT_SUGGESTIONS_ENABLED` for enabling suggestions


codewiththeflowxyz (approved):
I think this is a good approach!

PR #40, opened 2025-01-30
Fix issue with Trade widget

codewiththeflowxyz:
Fix crash in the trade widget.

Cow prices were strings like "0.9996915791616805" and when we tried to convert to a BigInt the hook would crash.

image
image

### How to test it
Since the error is triggered when BA Labs prices endpoint fails or return undefined for the sell or buy price, the best way to replicate the issue is by blocking that endpoint in your browser. You can test this in prod and compare with the deployment with the fix.

image


inclinedmind (approved):
tested it out, seems to work well

PR #46, opened 2025-01-31
Fix source and target tokens in url for sUSDS in the Trade widgets

codewiththeflowxyz:
Source and Target tokens in the search params didn't work for sUSDS in the Trade widgets (mainnet and Base) due to the mix of lower and upper case in the symbol.

It also prevents a situation where USDC could be both origin and target tokens when USDC was set as target_token in the url.


bitsforpeace (approved):
Works well for me, nice job

iceanddust (approved):

PR #51, opened 2025-02-01
fix issue with dialog and pasing empty string to img src

generatorpoint:
### What does this PR do?
Fixes some errors that were filling the console


packages/widgets/src/widgets/BalancesWidget/components/AssetBalance.tsx
```diff
@@ -50,7 +50,7 @@ export const AssetBalance = ({
           <Link to={action?.actionUrl || ''} hidden={!shouldShowAction} className="w-full">
             <div className="flex items-center justify-between">
               <div className="flex items-center space-x-2">
-                <img src={action?.image || ''} alt={`${symbol}-suggested-action`} className="h-[60px]" />
+                <img src={action?.image} alt={`${symbol}-suggested-action`} className="h-[60px]" />
```
generatorpoint, line 53: fixes `ERROR: An empty string ("") was passed to the src attribute.` 

this won't show if `shouldShowAction` is false

codewiththeflowxyz (approved):

PR #54, opened 2025-02-03
Update lint-staged, remove micromatch 4.0.5

bitsforpeace:
By updating lint-staged, we remove dep `micromatch@4.0.5` which should resolve dependabot alert [3](https://github.com/jetstreamgg/tarmac/security/dependabot/3)


iceanddust (approved):
LGTM, but looks like we also need to update it in the root package.json https://github.com/jetstreamgg/tarmac/blob/development/package.json#L81

PR #89, opened 2025-02-11
Link webapp to packages source files in dev mode for faster HMR

iceanddust:
While researching about ways to improve the DX and HMR speed in dev mode, I came across this official Vite [example repo](https://github.com/vitejs/vite-ts-monorepo-rfc) that implemented source file aliasing in the consumer application of a monorepo. I also received a very similar suggestion from ChatGPT, so I decided to try it out. It looks like this approach improves the HMR speed considerably in dev mode and also the cold start of the app locally.

Ideally, if implemented correctly, this should not produce any changes to the deployed versions of the app (Vercel, TechOps) but only apply to the dev version

### What does this PR do?
- Leverage Vite's path aliasing so the packages imports (utils, hooks and widgets) are fetched from their `./src` folder in dev mode, removing the need of waiting for a package rebuild on changes
- Add Tailwind code scanning for the widgets package in dev mode so we also have support for styles HMR with this approach
- Update widgets package's CSS file export so it is compatible with both dev and prod versions
- Rename widgets package's aliases from `@/*` to `@widgets/*` to avoid collision with webapp aliases in this new dev mode. This is where most of the code changes in this PR are coming from, unfortunately I couldn't find any workarounds for this
- Add path aliases to all `tsconfig.json` files (except for integration-ui app) so the editor can infer types and imports from Typescript files instead of built files
- Add a separate `tsconfig.production.json` file to the webapp, so the `typecheck` command gets the correct references from the built packages

TODO:
- [x] Check compatibility of Webapp and Widget styles
- [x] If we add files, or change types in one package, the other packages or webapp won't know about this until the package is rebuilt given our current configuration. How can we work around this? We could use different tsconfig files for dev mode, or explore the other approach of using Source export conditions (seems like Spark does this)

### Testing steps:
- Run this branch locally
- Install packages
- Run dev server with `pnpm dev`, the command now only runs the webapp
- Make changes to hooks or widgets package, they should automatically reflect
- Try to add new files to one of the packages and import them from the webapp, it should work without needing to build the packages again (don't forget to add the export to the corresponding `index.ts` file)


generatorpoint (approved):

codewiththeflowxyz (approved):
Really fast!

PR #95, opened 2025-02-12
Rename Tab search params actions to Flow

codewiththeflowxyz:
We were using the query param Tab with possible values of 'left', 'right' or undefined. We replace them with Flow and the possible widget flow values, like 'supply' or 'withdraw'.

- Balances
- Rewards: renamed but not handled yet, as we will do it in a separate PR among the other query params and external state.
- Savings
- Upgrade
- Trade: no need as it has only one flow and no tabs
- Seal: not managed


apps/webapp/src/modules/balances/components/BalancesWidgetPane.tsx
```diff
@@ -0,0 +1,31 @@
+import { WidgetStateChangeParams, SavingsFlow, BalancesWidget } from '@jetstreamgg/widgets';
+import { useSearchParams } from 'react-router-dom';
+import { SharedProps } from '@/modules/app/types/Widgets';
+import { QueryParams } from '@/lib/constants';
+import { BalancesWidgetProps } from 'node_modules/@jetstreamgg/widgets/dist/widgets/BalancesWidget';
```
bitsforpeace, line 5: Import from node_modules?

bitsforpeace (commented):
I'm noticing in the Balances widget that if you change tabs, the flow param updates, but Savings/Upgrade/Seal widgets do not have the same behavior. The flow URL param works when it loads but doesn't update when you change tabs by clicking. Should they all behave like the Balances widget?

codewiththeflowxyz:
> I'm noticing in the Balances widget that if you change tabs, the flow param updates, but Savings/Upgrade/Seal widgets do not have the same behavior. The flow URL param works when it loads but doesn't update when you change tabs by clicking. Should they all behave like the Balances widget?

I works for me in Savings and Upgrade. I didn't manage it in Seal though.

bitsforpeace (approved):
Savings still only seems to work when connected. Looks good otherwise. I was wondering if we want to do something for Seal positions, e.g. "unseal on position 1", in which case we would need the flow param there too. We could add it later though.

codewiththeflowxyz:
> Savings still only seems to work when connected. Looks good otherwise. I was wondering if we want to do something for Seal positions, e.g. "unseal on position 1", in which case we would need the flow param there too. We could add it later though.

I've just pushed Savings now. For Rewards and Seal I was planning to do it in their own PR as they probably need work on the other query params too.

PR #111, opened 2025-02-17
Add ConfirmationWarningRow which will replace the confirmation modal

codewiththeflowxyz:
- Adds speedbumps to the chatbot suggested actions when they modify the state: amounts or tokens
- Dynamically selects the disclaimer based on the selected intent

https://github.com/user-attachments/assets/c8705985-35f0-4f96-8ee4-04007c86cb7a


iceanddust (approved):

apps/webapp/src/modules/chat/components/ChatIntentsRow.tsx
```diff
@@ -43,22 +48,25 @@ const IntentRow = ({ intent }: IntentRowProps) => {
...
+
+        if (!hasShownIntent(intent) && modifiesState) {
+          setConfirmationWarningOpened(true);
           setSelectedIntent(intent);
         } else {
           setChatHistory(prev => [...prev, intentSelectedMessage(intent)]);
           navigate(intentUrl);
         }
       }}
       onMouseEnter={() => {
```
bitsforpeace, line 67: Do we need this prop?

bitsforpeace (approved):
Looking good, worked well during testing

PR #116, opened 2025-02-17
Fixes for QA findings

iceanddust:
### What does this PR do?
- Add reward name in rewards statistics section
- Remove redundant "rewarded" word in the total rewarded card
- Fix Radix's DialogTitle error
- Fix the hydration errors

Style updates:
- Add missing background, hover, and active states to some modal buttons
- Fix background colors of Select Delegate and Select Reward cards in Seal module
- Fix the hover state background color for the balances widget history items (previously it was transparent)
- Remove increment and decrement buttons rendered in the token inputs (these were introduced with the Tailwind v4 update)
- Update the background color of the TokenInput for USDS from purple to orange


generatorpoint (approved):
nice clean console

PR #127, opened 2025-02-20
Handle Upgrade widget params

codewiththeflowxyz:
## Upgrade
- Handle Upgrade widget query params
- Update query params based on widget state

## L2 Savings
- Prevent that a non allowed token in the url set the token in the widget. For instance if you sat USDT it would appear as the selected token even it's only possible to use USDC and USDS. 

## TODOs
- [x] Remove testing buttons once approved and before merging


iceanddust (approved):

bitsforpeace (approved):
LGTM, just a reminder to remove the testing changes

PR #132, opened 2025-02-21
Fix opacity in StepIndicator

codewiththeflowxyz:
### After
Tailwind v4 opacity applied

https://github.com/user-attachments/assets/e3ec05a1-50e6-40e5-9492-cbe8b7b04253

### Before
No opacity applied

https://github.com/user-attachments/assets/70ff8de8-b434-44b1-8466-c8a81efaf96c


bitsforpeace (approved):
Good catch

PR #153, opened 2025-03-04
Reset widget flows when clicking chatbot actions

codewiththeflowxyz:
Adds a new prop to reset the widget flow

NOTE: 
- ~~When the wallet extension is waiting for user confirmation the network switching doesn't work and the UI gets frozen.~~ Chatbot buttons are now disabled when a transaction is waiting for user interaction.
- Some chatbot buttons are included for testing purposes, they need to be removed before merging the PR.


apps/webapp/src/modules/chat/components/ChatIntentsRow.tsx
```diff
@@ -2,28 +2,92 @@ import { Button } from '@/components/ui/button';
...
+      intent_description: 'Start Seal flow',
+      url: '/?network=ethereum&widget=seal&reset=true',
+      intent_id: 'seal'
+    }
+  ];
+};
+
 export const ChatIntentsRow = ({ intents }: ChatIntentsRowProps) => {
-  const { shouldShowConfirmationWarning } = useChatContext();
+  console.log('🚀 ~ ChatIntentsRow ~ intents:', intents);
```
bitsforpeace, line 75: Do we want to remove these logs or keep them for now?

bitsforpeace (approved):

apps/webapp/src/modules/chat/components/ChatIntentsRow.tsx
codewiththeflowxyz: I'll remove them before merging the pr.

PR #161, opened 2025-03-06
fix issues with tailwind classes not being included from widgets, inc…

generatorpoint:
…lude other updates from dependabot

### What does this PR do?
- moves the widgets import for tailwind into the globals css file to fix an issue with classes not being applied

TODO:
- [x] fix an issue with `Link` component in the widgets package related to `react-router-dom`


codewiththeflowxyz (approved):
Tested it and the styles look good 🎨 👨‍🎨

PR #163, opened 2025-03-07
feature/migration Seal engine migration

codewiththeflowxyz:
Very basic PR so far, adds a route to a new page `/seal-engine` that renders the Seal widget. 

Needs layout/ui work.

image

image


bitsforpeace (commented):
Seems unrelated, but I noticed the wallet address is missing in the header. Do you know why?

PR #173, opened 2025-03-14
Remove open position

codewiththeflowxyz:
- Disable new Seal positions at widget level when user has one or more positions 
- Disable new Seal position at widget level when user has no positions (the Open new position flow is replaced with an informative screen)
- Disable new Seal position at webapp level. Even though the widget doesn't allow to create new positions we show a different messages and we don't render the widget in the webapp is the user has no position.

# New position button disabled from the manage flow
image

# New position flow removed at **widget** level
image

# New position disabled at **webapp** level
image


packages/widgets/src/widgets/SealModuleWidget/index.tsx
```diff
@@ -459,11 +459,13 @@ function SealModuleWidgetWrapped({
       } else if (widgetState.flow === SealFlow.MANAGE && currentStep === SealStep.SUMMARY) {
         setButtonText(t`Confirm`);
       } else if (shouldOpenFromWidgetButton) {
-        setButtonText(t`Open a new position`);
+        setButtonText(t`New positions disabled`);
```
bitsforpeace, line 462: Do we need to run messages for this update?

bitsforpeace (approved):
LGTM but think this requires an i18n update

packages/widgets/src/widgets/SealModuleWidget/index.tsx
codewiththeflowxyz: Good catch

PR #189, opened 2025-03-24
Remove activation checkbox and about page from the Stake widget

codewiththeflowxyz:
- Removes About page
- Removes exit fee logic


iceanddust (approved):
LGTM

PR #194, opened 2025-03-28
Update EthFlow contract address in Mainnet

iceanddust:
### What does this PR do?
- Update the EthFlow contract address to the new deployment in Mainnet per the CoWswap team's request
- Comment the CLE contract in the Wagmi config as it was causing issues with the Wagmi CLI. We are currently not using neither the contract nor the ABI in the code

### DIFF
--- a/packages/hooks/src/contracts.ts
+++ b/packages/hooks/src/contracts.ts
@@ -106,7 +106,9 @@ export const contracts: { name: string; address: Record<ChainId, `0x${string}`> 
   {
     name: 'ethFlow',
     address: {
-      [mainnet.id]: '0x40A50cf069e992AA4536211B23F286eF88752187',
+      [mainnet.id]: '0xba3cb449bd2b4adddbc894d8697f5170800eadec',
+      // TODO: Update the Tenderly address to the mainnet address
+      // if we use a testnet forked after March 28, 2025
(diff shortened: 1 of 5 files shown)


codewiththeflowxyz (approved):

inclinedmind (approved):

packages/hooks/src/contracts.ts
```diff
@@ -106,7 +106,9 @@ export const contracts: { name: string; address: Record<ChainId, `0x${string}`>
   {
     name: 'ethFlow',
     address: {
-      [mainnet.id]: '0x40A50cf069e992AA4536211B23F286eF88752187',
+      [mainnet.id]: '0xba3cb449bd2b4adddbc894d8697f5170800eadec',
+      // TODO: Update the Tenderly address to the mainnet address
+      // if we use a testnet forked after March 28, 2025
```
generatorpoint, line 111: would this affect e2e tests?

generatorpoint (approved):
lgtm

packages/hooks/src/contracts.ts
iceanddust: No because the parent fork is still a child of the Sep_30 testnet. Once we update the Sep_30 testnet for a new one we may need to update this contract address

PR #196, opened 2025-03-31
Change Migrate back to Upgrade

codewiththeflowxyz:
We previously changed the Upgrade term to Migrate, we are reverting this change.


inclinedmind (commented):
Looks like one more run of messages is needed
Screenshot 2025-03-31 at 9 45 36 AM

codewiththeflowxyz:
> Looks like one more run of messages is needed Screenshot 2025-03-31 at 9 45 36 AM

Good catch, I've just pushed those changes.

adamgoth (approved):

PR #200, opened 2025-04-01
Show soon label in the Rewards tab for L2s

codewiththeflowxyz:
image

### DIFF
--- a/apps/webapp/src/modules/balances/components/BalancesModuleShowcase.tsx
+++ b/apps/webapp/src/modules/balances/components/BalancesModuleShowcase.tsx
@@ -3,14 +3,13 @@ import { HStack } from '@/modules/layout/components/HStack';
 import { t } from '@lingui/core/macro';
 import { ModuleCard } from '@/modules/balances/components/ModuleCard';
 import { useChainId } from 'wagmi';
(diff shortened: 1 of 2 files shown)


apps/webapp/src/modules/balances/components/BalancesModuleShowcase.tsx
```diff
@@ -22,6 +22,7 @@ export function BalancesModuleShowcase() {
             title={t`Access rewards without giving up control`}
             className="bg-sky-blue"
             notAvailable={isBase || isArbitrum}
+            soon={isBase || isArbitrum}
```
bitsforpeace, line 25: Minor thing but I think we have `isL2ChainId` which would be better to use here and above on `notAvailable` (maybe there's other places to it would make sense to update too)

bitsforpeace (approved):

iceanddust (approved):

apps/webapp/src/modules/balances/components/BalancesModuleShowcase.tsx
codewiththeflowxyz: Good catch, I checked other places in the code and we don't need to replace it there.

PR #208, opened 2025-04-07
Add details pane to Seal migration page

codewiththeflowxyz:
image

image


apps/webapp/src/modules/app/components/SealMigrationWidgetPane.tsx
```diff
@@ -0,0 +1,180 @@
...
+                <ArrowLeft className="self-center" />
+                <Heading tag="h3" variant="small" className="text-textSecondary">
+                  Back to Home
+                </Heading>
+              </HStack>
+            </Link>
+          )}
+          {isL2 ? (
+            <div className="text-center">
+              <Heading variant="large">Connect to Mainnet</Heading>
```
bitsforpeace, line 142: Should we leave the header as Seal Engine so users still know where they are, then just have a button below for "Connect to Mainnet" instead of the dropdown?

bitsforpeace: Also, I think we could remove the network selector in the widget header, like how it's not shown for L1 only widgets. Thoughts?

bitsforpeace (approved):
Minor thing but I noticed if you don't have a connected wallet, we show the new position UI. Since this is being deprecated, we should probably show the same deprecation page but the bottom text could say "Connect a wallet to view your positions" or something like that.

It could be done in a separate PR.

codewiththeflowxyz:
Good catch

PR #214, opened 2025-04-09
Add log error when execute is called with no simulation data

iceanddust:
### What does this PR do?
Previously we checked if the `simulationData.request` was defined in order to call `writeContract()`, but if it wasn't defined, the function would return silently, making debugging errors more difficult. This PR adds an error log when `useWriteContractFlow.execute()` is called while `simulationData.request` is undefined

### DIFF
--- a/packages/hooks/src/shared/useWriteContractFlow.ts
+++ b/packages/hooks/src/shared/useWriteContractFlow.ts
@@ -125,6 +125,13 @@ export function useWriteContractFlow<
     execute: () => {
       if (simulationData?.request) {
         writeContract(simulationData.request);
+      } else {
+        console.log(`ERROR: the contract interaction was triggered before the call was ready.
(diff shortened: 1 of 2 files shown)


packages/hooks/src/shared/useWriteContractFlow.ts
```diff
@@ -125,6 +125,14 @@ export function useWriteContractFlow<
     execute: () => {
       if (simulationData?.request) {
         writeContract(simulationData.request);
+      } else {
+        console.log(`ERROR: the contract interaction was triggered before the call was ready.
```
generatorpoint, line 129: lgtm, maybe `console.warn` here for visibility?

generatorpoint (approved):

codewiththeflowxyz (approved):

PR #218, opened 2025-04-16
Test/parallelruns

0x2tester:
Summary 
Each worker gets its own address from getTestWalletAddress 
Initial funding happens once per worker
Sets balance for specific address
Verifies balance after setting
Takes snapshot after funding

Maintains list of test addresses received from `eth_accounts` response
Maps worker indices to addresses
Used consistently across:
Initial funding
RPC mocks
Balance operations

Flow:
Worker starts → Gets unique address
Initial funding → Sets balances for that address
Snapshot taken → Preserves funded state
During tests:
RPC calls return same address (as per worker index)
Balance operations target same address
All operations verify the address matches the worker


codewiththeflowxyz (approved):
LGTM!

Great job.

iceanddust (approved):
Nice, LGTM

PR #227, opened 2025-04-25
Wire up stake module

iceanddust:
### What does this PR do?
- Update hooks used in the stake module, as some were still using old Seal hooks
- Update address in the stake hooks to use the correct staking module address

### Testing steps:
- Attempt to perform a full new open position flow and a manage position flow

### DIFF
--- a/apps/webapp/vite.config.ts
+++ b/apps/webapp/vite.config.ts
@@ -51,6 +51,7 @@ export default ({ mode }: { mode: modeEnum }) => {
       https://safe-transaction-arbitrum.safe.global
       https://safe-transaction-sepolia.safe.global
       https://vote.makerdao.com
+      https://vote.sky.money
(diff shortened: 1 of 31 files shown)


codewiththeflowxyz (commented):
Overall the code looks good but I noticed some small issues, not necessarily introduced in this PR:

- The SKY amount in position overview is wrong as it's multiplied by 24000 again.
- Rewards and delegate don't appear to be preselected in the manage flow even when they've been saved in the previous step
- The deployment seems to throw an error, maybe a missing env?
- The Stake e2e is failing, probably due the deployment error above

iceanddust:
Thank you for pointing out those, I'll work on fixing them

codewiththeflowxyz (approved):
Nice! Working great.

generatorpoint (approved):
great, nice we don't have to do that math.conversion anymore

PR #231, opened 2025-04-29
Update testnet used for tests and fix broken tests after the state update

iceanddust:
### What does this PR do?
- Update parent testnet used for test forks
- Fix the stake multicall test
- Update test addresses in e2e tests and hooks tests
- Remove Seal `openLockFreeSky` and Stake `openLockFreeMkr` test files since locking SKY in Seal or MKR in Stake is not possible
- Remove SKY to MKR revert path from tests since reverting SKY is not possible


codewiththeflowxyz (approved):
Good to see all green again ✅ :)

generatorpoint (approved):

PR #232, opened 2025-04-30
Show selected reward and delegate in the Stake flow

codewiththeflowxyz:
Display the already selected reward and delegate in the Stake widget during the wizard flow
Apply the same change to Seal


iceanddust (approved):

generatorpoint (approved):
good solution

PR #239, opened 2025-05-05
Add test in seal migration flow for when the old engine is already hoped

codewiththeflowxyz:
- Disable next button if `migrate` is not prepared where resuming migration after approving migration contract
- Add test in seal migration flow for when the old engine is already hoped, so it starts directly in the migration step, skipping the approval.

### DIFF
--- a/packages/widgets/src/widgets/SealModuleWidget/index.tsx
+++ b/packages/widgets/src/widgets/SealModuleWidget/index.tsx
@@ -330,7 +330,7 @@ function SealModuleWidgetWrapped({
       console.log(error);
     },
     // TODO: criteria should be old hope, new hope and new index
-    enabled: !!newStakeUrn?.urnIndex
+    enabled: newStakeUrn?.urnIndex !== undefined
   });
 
   const lockMkrApprove = useSaMkrApprove({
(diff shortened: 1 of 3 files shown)


packages/widgets/src/widgets/SealModuleWidget/index.tsx
```diff
@@ -751,6 +751,13 @@ function SealModuleWidgetWrapped({
           currentStep === SealStep.HOPE_OLD &&
           txStatus === TxStatus.SUCCESS &&
           !migrate.prepared) ||
+        // Disable next button if `migrate` is not prepared where resuming migration after approving migration contract
+        (widgetState.flow === SealFlow.MIGRATE &&
+          currentStep === SealStep.ABOUT &&
+          txStatus === TxStatus.IDLE &&
+          !needsNewUrnAuth &&
+          !needsOldUrnAuth &&
```
iceanddust, line 759: Do we need to add `needsNewUrnAuth` and `needsOldUrnAuth` to the dependency array of this `useEffect`?

iceanddust (approved):
LGTM! Just left a question about the dependency array of the useEffect

PR #244, opened 2025-05-08
Migration index selector

codewiththeflowxyz:


generatorpoint:
Great work so far! One suggestion, maybe if there are not staking urns open we can auto select "create new urn" in the dropdown

packages/widgets/src/widgets/SealModuleWidget/components/MigrateAbout.tsx
```diff
@@ -235,6 +296,42 @@ const InfoCard = ({
             <Text variant="medium" className="mb-1 font-medium">
               {title}
             </Text>
+            {showSelector && currentStakeUrnIndex !== undefined && (
+              <VStack gap={2} className="my-4">
+                <Text variant="small" className="text-textSecondary">
+                  Select an existing Stake position to migrate to, or create a new one:
+                </Text>
+                <Select
```
iceanddust, line 304: Maybe for this select we can reuse the styles of the `TokenInput` dropdown for consistency. I think it's these, but we'll have to try `bg-container rounded-[20px] border-0 p-2 pr-0 pt-5 backdrop-blur-[50px]`

iceanddust (approved):
LGTM, I just left a comment about the styles of the new Select component

packages/widgets/src/widgets/SealModuleWidget/components/MigrateAbout.tsx
generatorpoint: also, I think we need cursor: pointer for the select items

generatorpoint (approved):

PR #255, opened 2025-05-13
Staking subgraph updates

inclinedmind:
- Update useStakingHistory hook to fetch the correct staking events from the subgraph (instead of seal events)
- Also fixes a bug where we were showing all events in the history for position 0


apps/webapp/src/modules/stake/components/StakeHistory.tsx
```diff
@@ -38,32 +36,30 @@ const mapTypeEnumToColumn = (type: TransactionTypeEnum) => {
...
   }
 };
 
-const highlightedEvents = [TransactionTypeEnum.SEAL, TransactionTypeEnum.SEAL_SKY, TransactionTypeEnum.REPAY];
+const highlightedEvents = [TransactionTypeEnum.STAKE, TransactionTypeEnum.STAKE_REPAY];
 
 export function StakeHistory({ index }: { index?: number }) {
-  const { data: sealHistory, isLoading: sealHistoryLoading, error } = useSealHistory({ index });
+  const { data: stakeHistory, isLoading: stakeHistoryLoading, error } = useStakeHistory({ index });
+  console.log('stakeHistory', stakeHistory);
```
generatorpoint, line 55: probably want to remove this log

apps/webapp/src/modules/stake/components/StakeHistory.tsx
```diff
@@ -9,26 +9,24 @@ import { HistoryTable } from '@/modules/ui/components/historyTable/HistoryTable'
 // TODO: This needs to be updated once we have the stake history coming from hooks
```
generatorpoint, line 9: can we remove this TODO now, or does it apply to something else?

generatorpoint (approved):

PR #260, opened 2025-05-14
Round debt up to the nearest whole number

codewiththeflowxyz:
Note: For the images the numbers have been hardcoded 

### 10.1
image

### 11.00000000000000000001
image

### 15 or 15.0
image


packages/widgets/src/widgets/StakeModuleWidget/components/Repay.tsx
```diff
@@ -293,9 +293,17 @@ export const Repay = ({ isConnectedAndEnabled }: { isConnectedAndEnabled: boolea
     unit: getTokenDecimals(usds, chainId)
   })}`;
 
-  const formattedDebtValue = `${formatBigInt(existingVault?.debtValue || 0n, {
+  const formattedRoundedDebtValue = formatBigInt(existingVault?.debtValue || 0n, {
```
bitsforpeace, line 296: We probably want to make this a helper, maybe in one of the packages, because I think we need to use the same formatting in the detail pane. Thoughts?

Screenshot 2025-05-15 at 2 08 46 PM

codewiththeflowxyz: I've made this as a helper function, but maybe there we can leave it as is because it's used in BaseBalanceCard and maybe affects other card types.

PR #271, opened 2025-05-15
Fix the wrong token selected in revert flow

iceanddust:
### What does this PR do?
- Add USDS as a valid source token for the upgrade widget to avoid the query parameter being wiped, and the revert flow to be locked
- Add a check when user clicks on the upgrade widget tabs to avoid updating the widget state if the user clicks on the same tab
- Prevent removing the `source_token` URL parameter in revert flows. Which was causing the flow to break after approving the token
- Update the revert flow token input label since now it's not possible to select revert tokens

### Testing steps:
- Go to the upgrade widget
- Switch between the upgrade and revert tabs, the tokens should update correctly
- Attempt to perform an approve + revert flow


inclinedmind (approved):
Looks good, we should also remove the text that says "choose a token to revert" since there's no choice to make

codewiththeflowxyz (approved):

PR #312, opened 2025-05-30
Add Unichain and Optimism Safe wallet prefixes

iceanddust:
### What does this PR do?
Add the wallet prefixes for Safe wallets in Unichain and Optimism. These are used to generate the Safe transaction URL


inclinedmind (approved):

generatorpoint (approved):
good catch

PR #333, opened 2025-06-03
pass a chain value to the fork script

generatorpoint:


inclinedmind (approved):
Looks good 👍 One question - why do we need all the chains for the reward-2 spec?

generatorpoint:
> Looks good 👍 One question - why do we need all the chains for the reward-2 spec?

I noticed the test goes to the balances tab at one point and there it makes requests to fetch balances from all the other chains

PR #335, opened 2025-06-03
Add batch transactions to L2 savings widget

iceanddust:
### What does this PR do?
Adds support for batch transactions in the L2 savings widget (Metamask currently only supports Base), to work in parallel with common transactions

### Testing steps:
- Connect with the shared testing account in Base and navigate to the L2 savings widget
- Attempt to supply with an amount that requires allowance
- The widget button should automatically trigger the 'Upgrade' step, without needing the approval step before
- Trigger the transaction and check in Metamask that the transaction populated is indeed a batch transaction
- Confirm the tx in the wallet and wait for it to be executed. The webapp should display the status accordingly


inclinedmind (approved):
looks good, tested out both batched and non-batched transactions in the widgets and they worked well

bitsforpeace (approved):
Works great for me

bitsforpeace (commented):
Only issue is MM seems to warn to reject the tx when withdrawing for Savings, but not sure why because everything happens as expected

Screenshot 2025-06-26 at 9 34 54 AM

PR #345, opened 2025-06-05
Fix unnecessary extra rerenders for the widget inputs

iceanddust:
### What does this PR do?
- Only update inputs after external state has changed if the amount in the search params is different than the current input amount
- Clean up the timeout function
- Update the external state based on the debounced origin amount instead of on every input change
- Remove the `useNotifyWidgetState` call which was causing double rerender cycles
- Update the `isAmountWaitingForDebounce` variable to only wait for debounce on a single token: the one that was last updated, as the other token should be updated automatically by the quote hooks and shouldn't wait for debounce in that case


codewiththeflowxyz:
Should this PR be against `development` instead?

PR #360, opened 2025-06-11
Add batch transaction tests

iceanddust:


generatorpoint:
great, works for me after applying the local patches you suggested. It's worth noting the `Base trade tests › Batch - trade usdc to usds` test passed even before applying the patch. Are we able to verify that the test transaction was performed as EIP 7702 to rule out false positives?

PR #362, opened 2025-06-11
Don't allow users to switch networks from the ChainModal when on Safe

codewiththeflowxyz:
### Safe{Wallet}
image

### Normal

image


iceanddust (approved):
Nice improvement, LGTM

PR #369, opened 2025-06-13
Prevent details pane to remain on screen when going from mobile bpi to desktop

codewiththeflowxyz:
When the user navigates from a mobile breakpoint to a desktop breakpoint we were rendering 2 Details pane at the same time. One in the widgets pane (like the mobile view) and one in the details pane.

image


iceanddust (approved):
LGTM

PR #397, opened 2025-06-24
Add additional ignore patterns in PR description script

bitsforpeace:


iceanddust (approved):
LGTM, do we maybe also want to add the generated file `packages/hooks/src/generated.ts`? It does contain important code changes, but I think the same changes can be tracked by the `contracts.ts` file

bitsforpeace:
> LGTM, do we maybe also want to add the generated file `packages/hooks/src/generated.ts`? It does contain important code changes, but I think the same changes can be tracked by the `contracts.ts` file

I'll update it

PR #441, opened 2025-06-26
Batch toggle disabling functionality and read feature flag for initial state of `batchEnabled`

iceanddust:
### What does this PR do?

This PR improves the user experience for batch transactions by:

1.  Disabling the batch transaction toggle button when the user's connected wallet or chain does not support multicall functionality.
2.  Displaying an informative tooltip explaining why the toggle is disabled in this scenario.
3.  Use the `VITE_BATCH_TX_ENABLED` environment variable feature flag to set the initial state of `batchEnabled`, this allows to disable batch transactions in the widget as well.

### Testing steps:

1.  **Test Unsupported Wallet/Chain:**
    *   Connect a wallet or switch to a network that does not support batch transactions.
    *   Verify the batch transaction toggle (Zap icon) is disabled and cannot be clicked.
    *   Hover over the toggle and confirm the tooltip reads: `Your wallet or the chain you are connected to currently does not support bundled transactions`.

2.  **Test Supported Wallet/Chain:**
    *   Connect a wallet and network that supports batch transactions.
    *   Verify the toggle is enabled and can be switched on and off.
    *   Verify the tooltip correctly shows `Bundled transactions enabled` or `Bundled transactions disabled` as you interact with it.

3.  **Test Feature Flag:**
    *   Set the `VITE_BATCH_TX_ENABLED` environment variable to `false` and restart the application.
    *   Verify that batch transactions are disabled by default.
    *   If the connected chain supports it, enable the toggle, then refresh the page.
    *   Confirm that batching is once again disabled, showing the feature flag correctly overrides the user's saved setting.


generatorpoint (approved):

generatorpoint (approved):
LGTM

PR #526, opened 2025-07-08
always default to dai / usds when toggling tabs in upgrade widget

inclinedmind:
instead of using the same tokens already selected


codewiththeflowxyz (approved):
I tested it and it's working well.

PR #550, opened 2025-07-11
[Widgets Refactor]: Add `onMutate` callback to write hooks and move widget initial transaction updates to the callback

iceanddust:
Builds on #511, merge that one first

### What does this PR do?

This PR refactors transaction handling across all widgets by introducing a new `onMutate` callback to our write-hooks.

This `onMutate` callback is fired immediately when a transaction is initiated, but before the user accepts the transaction in their wallet, allowing us to centralize the UI logic for preparing the transaction view (e.g., switching screens, setting transaction status to `INITIALIZED`).

This change removes a significant amount of boilerplate code from the widget components (e.g., `approveOnClick`, `supplyOnClick` handlers) and simplifies the overall control flow, especially for multi-step transactions like approve-then-act.

Additionally, this PR also removes the code that would check if `nextOnClick()` was being called after a successful approval to trigger the next transaction, since now that happens automatically and is never triggered from this button.

### Testing steps:

1.  Navigate to each of the following widgets:
    *   L2 Savings
    *   L2 Trade
    *   Rewards
    *   Savings (L1)
    *   Stake
    *   Upgrade
2.  For each widget, perform the primary user flows (e.g., Supply, Withdraw, Trade, Claim, etc.).
3.  **Verify that upon clicking an action button (e.g., "Approve", "Supply", "Trade"), the UI immediately switches to the transaction screen *before* the wallet prompt appears.**
4.  Test the full lifecycle of transactions:
    *   Execute an approval, then ensure the subsequent action (e.g., Supply, Trade) is triggered correctly.
    *   Execute batch/multicall transactions where applicable.
    *   Confirm that the success screen appears and the "Next"/"Finish" buttons reset the widget state correctly.
    *   Reject a transaction in your wallet and verify that the error screen is shown and the "Try Again" button works as expected.
5.  Specifically test the "Withdraw All" functionality in the L2 Savings widget to ensure it works correctly.

PR #560, opened 2025-07-15
[Transaction flows refactor]: Unified transaction flow hooks

iceanddust:
Builds on #550, merge that one first

### What does this PR do?

This PR introduces a unified `useTransactionFlow` hook to abstract the handling of both batch and sequential (e.g., `approve` + action) transactions.

It adds a new `useSequentialTransactionFlow` hook that executes an array of transactions one-by-one, providing a fallback for wallets that do not support batching, or for when the user has set the preference to not use batch transactions.

All major transaction hooks and widgets (`L2Savings`, `L2Trade`, `Rewards`, `Savings`, `Stake`, `Upgrade`) have been refactored to use this new abstraction. This greatly simplifies the widget code by removing manual state management for separate approval and action flows, centralizing all transaction execution logic.

### Testing steps:

**1. With bundled transactions enabled:**
  - Go to any widget (e.g., Savings, Trade, Stake).
  - Perform an action that requires an allowance (e.g., supply for the first time).
  - **Expected:** You should see a single, batched transaction request.
  - Confirm the transaction and verify it succeeds.

**2. With bundled transactions disabled, or with a wallet that doesn't support the feature:**
  - Go to any widget.
  - Perform an action that requires an allowance.
  - **Expected:** You should be prompted to sign the `approve` transaction first, followed automatically by a prompt for the main action (e.g., `supply`).
  - Confirm both transactions and verify they succeed.

**3. With pre-existing allowance:**
  - Go to any widget where you already have sufficient token allowance.
  - Perform an action.
  - **Expected:** You should only be prompted to sign a single transaction for the main action, regardless of the wallet used.
   
**4. Tests the edge case scenarios:**
  - Cancel the approve transaction and attempt to retry it.
  - Cancel the 2nd transaction and attempt to retry it.
  - Cancel any of the transactions, go back and attempt to send a new transaction flow, or modify the inputs and try.


packages/hooks/src/savings/useBatchSavingsSupply.ts
```diff
@@ -5,7 +5,7 @@ import { sUsdsAddress, sUsdsImplementationAbi } from './useReadSavingsUsds';
 import { getWriteContractCall } from '../shared/getWriteContractCall';
 import { usdsAddress } from '../generated';
 import { Call, erc20Abi } from 'viem';
-import { useSendBatchTransactionFlow } from '../shared/useSendBatchTransactionFlow';
+import { useTransactionFlow } from '../shared/useTransactionFlow';
 
 export function useBatchSavingsSupply({
```
codewiththeflowxyz, line 10: Does it make sense to keep hooks like `useSavingsSupply` or `useStakeMulticall` if we have these `useBatch...` version? It looks like we're not using these in the repo anymore. I we can delete them maybe we should rename this `useBatchSavingsSupply` and the other hooks by deleting the `Batch` word from the name. wdyt?

codewiththeflowxyz (approved):
LGTM! Easy to understand and it abstracts the batch/sequential logic away making it easier to use in the widgets.

Left a comment that applies to all useBatchXYZ hooks.

packages/hooks/src/savings/useBatchSavingsSupply.ts
iceanddust: Yeah I've been thinking about this. It definitely doesn't make much sense to keep 2 hooks that do the same thing. Not sure if we should keep the old ones for a while for backwards compatibility, maybe not since the new hooks should be compatible with the old ones. In that sense yeah maybe renaming the batch hooks make sense. Or name them something like use[action]Flow, not sure

inclinedmind (approved):

iceanddust:
@coderabbbitai review

packages/hooks/src/shared/useTransactionFlow.ts
```diff
@@ -0,0 +1,56 @@
...
+    ...commonTransactionParameters,
+    enabled: enabled && !useBatch && !isLoadingCapabilities,
+    gcTime
+  });
+
+  // Use batch flow
+  const batchResults = useSendBatchTransactionFlow({
+    ...commonTransactionParameters,
+    enabled: enabled && useBatch && !isLoadingCapabilities
+  });
```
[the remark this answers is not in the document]
iceanddust: `gcTime` is relevant for the sequential flow hook because it performs a query using `useSimulateContract`, but the batch flow hook doesn't perform the simulation and directly writes to the contract

PR #574, opened 2025-07-18
Bundled action: upgrade and supply in Savings widget

iceanddust:
### What does this PR do?

This PR introduces a new transaction flow allowing users to supply DAI directly to the mainnet Savings module. This is achieved by bundling the required actions (DAI approval, upgrade to USDS, USDS approval, and supply) into a single transaction, and the feature is only available for bundled transaction flows.

A new `useBatchUpgradeAndSavingsSupply` hook encapsulates this logic. The Savings widget UI is updated to allow selecting DAI as an input, display a custom 4-step progress indicator for this new flow, warn users if their wallet doesn't support bundled transactions or if they have the feature disabled, and disable the widget button and the transaction if this is the case.

For practicality and ease of testing, the feature is enabled as a sequential transaction flow only in testnets.

### Testing steps:

1.  Navigate to the Savings widget on Mainnet.
2.  In the "Supply" tab, select "DAI" from the token dropdown menu.
3.  Enter an amount of DAI to supply and disable bundled transactions. Check that the warning appears.
4.  Proceed to the review screen.
5.  Confirm the 4-step transaction plan is displayed (Approve DAI, Upgrade, Approve USDS, Supply).
6.  Execute the transaction and check that it is a single, bundled transaction containing 2-4 actions depending on the user allowances.
7.  Cancel the transaction and repeat steps 1-6 in the Tenderly testnet. The transactions should be sent individually now.
9.  After success, verify that your DAI balance has decreased and your sUSDS (savings) balance has increased correctly.
10.  Test deep linking by navigating to `/savings?widget=savings&sourceToken=DAI&inputAmount=100`. The widget should be pre-filled with 100 DAI.

https://github.com/user-attachments/assets/606d4d0b-7ef5-4127-825a-4458f11e6944

### DIFF
--- a/packages/widgets/src/shared/components/ui/BundledTransactionWarning.tsx
+++ b/packages/widgets/src/shared/components/ui/BundledTransactionWarning.tsx
@@ -3,7 +3,7 @@ import { Text } from '@widgets/shared/components/ui/Typography';
 import { InfoTooltip } from '@widgets/shared/components/ui/tooltip/InfoTooltip';
 import { positionAnimations } from '@widgets/shared/animation/presets';
 import { useIsBatchSupported } from '@jetstreamgg/sky-hooks';
-import { useChainId } from 'wagmi';
+import { useChainId, useConnection } from 'wagmi';
 import { isTestnetId } from '@jetstreamgg/sky-utils';
 
 /**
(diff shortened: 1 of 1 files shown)


inclinedmind (commented):
Screenshot 2025-08-20 at 3 06 29 PM

I think this should say `How much DAI would you like to supply` here.

Also, if I select DAI in the supply dropdown, and then go to the withdraw, tab, it will incorrectly show DAI instead of USDS in a few places
Screenshot 2025-08-20 at 3 19 52 PM
Screenshot 2025-08-20 at 3 19 56 PM

PR #589, opened 2025-07-22
QA items

iceanddust:
### What does this PR do?
- Prevent token input popover from colliding with edge of viewport:
On small height screens, the token input popover would render and be cut by the edge of the viewport. Setting the `avoidCollisions` property to `true` lets the popover detect edges render the popover content above if there's not enough space below.
- Fix an issue where using the `formatNumber` function with the `maxDecimals` set to something larger than `SMALL_NUM_DECIMALS (4)` would cause small numbers to be formatted as `<0%`
- Fix issue that prevented popovers to be scrolled through when rendered on top of modals

### Testing steps:
- Decrease the viewport screen enough so the popover doesn't have enough space to render below and check that it correctly renders above.
- Select different tokens and check that the behavior is the expected one.
- There used to be a previous issue where this property would cause the popover flashing. Check that this is not the case anymore.
For the popover scroll issue: 
- Navigate to the stake page and go to the open position flow
- At the top of the widget, click on the `(i)` icon next to the `1/4 STEPS` to open the modal
- Next, click on the `(i)` icon in the (3) delegate section to open the popover
- Attempt to scroll through the text using the mouse wheel, it should be possible now

### DIFF
--- a/packages/utils/src/formatValue.ts
+++ b/packages/utils/src/formatValue.ts
@@ -67,7 +67,8 @@ export function formatBigInt(amount: bigint, options?: FormatOptions): string {
 
 export function formatNumber(amount: number, options?: FormatOptions): string {
   const absAmount = Math.abs(amount);
-  const smallestNumber = 1 / Math.pow(10, SMALL_NUM_DECIMALS); //0.0001 if SMALL_NUM_DECIMALS is 4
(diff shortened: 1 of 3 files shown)


inclinedmind (approved):
all looks good to me 👍

codewiththeflowxyz (approved):
It works well 👍

PR #592, opened 2025-07-23
export widget tooltips from a data folder

generatorpoint:

### DIFF
--- a/packages/widgets/src/data/tooltips/index.ts
+++ b/packages/widgets/src/data/tooltips/index.ts
@@ -1,0 +1,128 @@
+export interface Tooltip {
+  id: string;
+  title: string;
+  tooltip: string;
+}
+
+export const tooltips: Tooltip[] = [
+  // Stake Module tooltips
+  {
+    id: 'collateralization-ratio',
+    title: 'Collateralization ratio',
+    tooltip:
+      "The ratio between the value of collateral you've provided and the amount you've borrowed against that collateral."
+  },
+  {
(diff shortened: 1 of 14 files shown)


codewiththeflowxyz (approved):

packages/widgets/src/data/tooltips/index.ts
```diff
@@ -0,0 +1,128 @@
...
+  {
+    id: 'borrow-rate-seal',
+    title: 'Borrow rate',
+    tooltip:
+      'The Borrow Rate is determined by Sky Ecosystem Governance through a process of community-driven, decentralized onchain voting.'
+  }
+];
+
+// Helper function to get tooltip by ID
+export function getTooltipById(id: string): Tooltip | undefined {
```
iceanddust, line 126: I'm wondering if it'd be possible to type the `id` parameter based on the values of the `tooltip` array. That way, we could get autocompletion when calling `getTooltipById()`, making it less prone to typos.
Maybe something like this could work:
```diff
- export const tooltips: Tooltip[] = [...];
+ export const tooltips: Tooltip[] = [...] as const;

+ export type TooltipId = (typeof tooltips)[number]['id'];

- export function getTooltipById(id: string) {...}
+ export function getTooltipById(id: TooltipId) {...}
```

iceanddust (approved):
Nice, I like where this is going. I left an idea about the possibility to add autocompletion for tooltip IDs

PR #593, opened 2025-07-24
add stusds history hooks and components

inclinedmind:
Uses the savings deposit & withdraw icons for now, we can update if we decide to use separate icons


bitsforpeace (approved):
Looking good!

codewiththeflowxyz (approved):

PR #629, opened 2025-07-29
handle stusds balance module filtering

inclinedmind:
- always hide the stusds module card if the balance is 0
- show the correct data on non-mainnet networks
- filter out the card if we're only showing cards on a different network


codewiththeflowxyz (approved):
LGTM! I'm seeing the "USDS supplied to stUSDS" card correctly but not the stUSDS balance in the token list.

PR #632, opened 2025-07-30
Add available liquidity and total debt stats to the stusds details panel

inclinedmind:
Also, updated the layout to always flex-wrap on all screen sizes, so that you never need to horizontally scroll to see all the stats. If this looks good we can update the other detail pages to do the same.

Popover copy is still TODO.

Screenshot 2025-07-30 at 2 46 43 PM


iceanddust (approved):
LGTM, but it looks like the cards in the balances widget got broken, maybe because of the `h-full` addition to the stats card? Maybe this gets fixed by implementing the wrapping of the cards there too though

Image

codewiththeflowxyz (approved):

PR #670, opened 2025-08-12
Expert modules disclaimer card

iceanddust:
### What does this PR do?

It replaces the mandatory risk acknowledgment checkbox with a one-time, dismissible disclaimer banner.

Expert options are now enabled as soon as the disclaimer is displayed, rather than requiring an explicit user opt-in. This is supported by renaming the `expertRiskAcknowledged` state and related props to `expertRiskDisclaimerShown` to better reflect the new logic.

### Testing steps:

1.  Navigate to the Expert Modules section for the first time (or after clearing local storage).
2.  **Verify:** The new risk disclaimer banner is displayed.
3.  **Verify:** The "stUSDS" module card below the banner is **enabled** and clickable.
4.  Click the 'X' icon to dismiss the banner.
5.  **Verify:** The banner disappears.
6.  Refresh the page.
7.  **Verify:** The banner does **not** reappear and the "stUSDS" module remains enabled.
8.  (Optional) Clear `expert-risk-disclaimer-dismissed` from local storage and refresh to confirm the banner shows again.


codewiththeflowxyz:
image
the stUSDS card should be enabled because the disclaimer is rendered. Atm the first time you load the module the card is disabled, if you navigate to a different module and back the card will be enabled. I think this should happen in the first render.

codewiththeflowxyz (commented):
LGTM but there's a small thing, the card should be enabled when you first visit the module.

inclinedmind (approved):

PR #673, opened 2025-08-12
Move ConfigProvider to separate file to fix HMR

iceanddust:
### What does this PR do?

This PR fixes an issue that was causing HMR in dev mode to break. Whenever HMR was applied, the `ConfigContext` was reset to its default props, and they were never updated to the actual values (so even the update functions were a no-op).

### Testing steps:

1.  Run the webapp locally
2. Attempt to navigate through the widgets, it should work well
3. Attempt to perform a change in any webapp files and save (I tested it by adding a console.log() to the `SavingsWidgetPane` file). After saving, the app should automatically update without requiring a page reload. Widget navigation should still work fine and the selected widget should remain (so not redirect to the balances page).
4. Try to use the app normally


codewiththeflowxyz (approved):
Tested it out and it works well. LGTM! nice catch.

PR #688, opened 2025-08-14
Staging -> main

iceanddust:
### Changes included:
- #592 
- #669
- #530
- #679 (and fix trade header menu collisions)
- #673


apps/webapp/src/modules/app/components/DetailsPane.tsx
```diff
@@ -101,10 +102,16 @@ export const DetailsPane = ({ intent }: DetailsPaneProps) => {
                   <StakeDetailsPane />
                 </MotionDetailsWrapper>
               );
+            case Intent.SEAL_INTENT:
```
bitsforpeace, line 105: Anyone recall why we added this?

codewiththeflowxyz (approved):

apps/webapp/src/modules/app/components/DetailsPane.tsx
inclinedmind: without this we're just showing the balances details on the seal page

inclinedmind (dismissed):

generatorpoint (approved):

PR #714, opened 2025-08-27
useId in graph

inclinedmind:
1. When switching from Chart A to Chart B, both might briefly exist in the DOM during the transition
  3. Both charts would define a gradient with ID "gradientGreen", creating a temporary duplicate
  4. Safari might grab the wrong gradient reference during this transition period, showing black

  The useId fix ensures that even during these transition overlaps, each chart instance has its own unique gradient ID, preventing any reference confusion.

I only saw the issue on mobile safari.


generatorpoint (approved):
nice use of `useId`

PR #741, opened 2025-09-01
Refactor: Simplify getTokenDecimals usage by handling undefined tokens internally

codewiththeflowxyz:
This PR refactors the `getTokenDecimals` function to handle undefined/null tokens internally, eliminating the need for defensive ternary checks throughout the codebase. The function signature now explicitly accepts `Token | TokenForChain | undefined | null` and returns `defaultDecimals` if defined or a default value of 18 decimals when the token is undefined.

As a result of this change, we've removed 40+ redundant ternary checks across 15 files, simplifying patterns like `token ? getTokenDecimals(token, chainId) : 18` to just `getTokenDecimals(token, chainId)`. This makes the code cleaner, more maintainable, and ensures consistent handling of undefined tokens across the entire application.

### Core Function Update
- Updated `getTokenDecimals` in `/packages/hooks/src/tokens/tokens.constants.ts` to accept undefined/null tokens
- Function now returns default decimals (18) when token is undefined

### DIFF
--- a/packages/hooks/src/psm/usePreviewSwapExactIn.ts
+++ b/packages/hooks/src/psm/usePreviewSwapExactIn.ts
@@ -29,7 +29,7 @@ export const usePreviewSwapExactIn = (
     };
   }
   // use the correct decimals for the out token
-  const tokenDecimals = outToken ? getTokenDecimals(outToken, chainId) : 18;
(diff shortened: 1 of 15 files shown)


bitsforpeace (approved):
Nice, much cleaner

PR #756, opened 2025-09-02
Usdt reset allowance (into staging)

inclinedmind:
Replacement for https://github.com/jetstreamgg/tarmac/pull/731

What does this PR do?

  Adds batched USDT approval functionality to handle USDT's unique requirement of resetting allowance to 0 before setting a new amount.
  Includes a new useBatchUsdtApprove hook and UI toggle for bundled transactions.

I confirmed that it works on both sepolia and mainnet

  Testing steps:

  1. Navigate to the trade widget with USDT as the origin token
  2. Set a trade amount that requires increasing the current non-zero USDT allowance
  3. Verify the "Bundle transactions" toggle appears in the trade summary
  4. Test with toggle enabled: Should execute both reset (to 0) and approve calls in a single batch transaction
  5. Test with toggle disabled: Should show error message requiring bundled transactions to be enabled
  6. Verify normal ERC-20 tokens still use the standard approval flow without the toggle

Screenshot 2025-08-29 at 5 00 04 PM

Screenshot 2025-08-29 at 5 00 09 PM


iceanddust (approved):
Amazing, works really well!

bitsforpeace (approved):

PR #760, opened 2025-09-03
[multichain UX] track network across widgets

bitsforpeace:
## What does this PR do?

This PR implements widget-specific network tracking for multichain widgets in the Sky Protocol app. The feature addresses a UX issue where users lose their preferred network context when navigating between widgets.

### Problem Solved
Previously, when users navigated from a multichain widget (e.g., Savings on Base) to a mainnet-only widget (e.g., Upgrade) and then back to the original widget, they would remain on mainnet instead of returning to their preferred network (Base).

### Implementation Details

**Key Changes:**
- Extended `NetworkSwitchContext` to track network preferences per widget during the session
- Modified `WidgetNavigation` component to save/restore network states when switching between widgets
- Updated `ChainModal` to track manual network changes for the current widget

**Tracking Behavior:**
- **Trade & Savings widgets**: Each maintains its own network preference independently
  - When leaving: Current network is saved for that specific widget
  - When returning: Widget's previously saved network is restored
- **Balances widget**: Explicitly excluded from tracking - always stays on the current network
- **Mainnet-only widgets** (Upgrade, Rewards, Stake, Expert, Seal): Continue to force mainnet as before

**Session Scope:**
- Network preferences are stored in React state (session-only)
- No browser storage is used
- All preferences are cleared on page refresh

### Files Modified
- `/apps/webapp/src/modules/ui/context/NetworkSwitchContext.tsx` - Added widget network state management
- `/apps/webapp/src/modules/app/components/WidgetNavigation.tsx` - Implemented save/restore logic
- `/apps/webapp/src/modules/ui/components/ChainModal.tsx` - Added tracking for manual network changes

## Testing Steps

1. **Test independent widget tracking:**
   - Navigate to Trade widget and switch to Optimism
   - Navigate to Savings widget and switch to Base
   - Navigate to Stake widget (forced to mainnet)
   - Navigate back to Trade → **Should restore Optimism** (not Base from Savings)
   - Navigate to Savings → **Should restore Base**

2. **Test Balances widget exclusion:**
   - Navigate to Trade and set to Arbitrum
   - Navigate to Balances (stays on Arbitrum)
   - Manually switch to Base while in Balances
   - Navigate to Upgrade (forced to mainnet)
   - Navigate back to Balances → **Should stay on mainnet** (no restoration)
   - Navigate to Trade → **Should restore Arbitrum**

3. **Test mainnet-only to multichain transitions:**
   - Navigate to Savings and set to Base
   - Navigate to Upgrade (forced to mainnet)
   - Navigate back to Savings → **Should restore Base**

4. **Test multichain to multichain navigation:**
   - Navigate to Trade (no saved state, on mainnet)
   - Switch to Arbitrum
   - Navigate to Savings (no saved state) → **Should stay on Arbitrum**
   - Navigate back to Trade → **Should stay on Arbitrum** (same as saved)

5. **Test manual network switching updates:**
   - Navigate to Trade and switch to Optimism
   - Manually switch to Base using the network switcher
   - Navigate to Rewards (forced to mainnet)
   - Navigate back to Trade → **Should restore Base** (the manually selected network)

6. **Test session persistence:**
   - Set different networks for Trade (Optimism) and Savings (Base)
   - Navigate between various widgets to confirm states persist
   - Refresh the page
   - Navigate to Trade → **Should not restore** (cleared on refresh)
   - Navigate to Savings → **Should not restore** (cleared on refresh)

7. **Test edge cases:**
   - Rapidly switch between widgets to ensure no race conditions
   - Test with network switch failures (reject in wallet) - state should remain consistent
   - Navigate directly via URL changes to ensure tracking still works

### Expected Behavior Summary
- Trade and Savings each remember their own last-used network
- Balances never saves or restores, always uses current network  
- Network preferences are widget-specific, not shared between widgets
- Manual network changes update the saved state for the current widget
- All tracking is session-only and clears on page refresh

### DIFF
--- a/apps/webapp/src/modules/app/components/WidgetNavigation.tsx
+++ b/apps/webapp/src/modules/app/components/WidgetNavigation.tsx
@@ -10,7 +10,7 @@ import { cardAnimations } from '@/modules/ui/animation/presets';
 import { AnimationLabels } from '@/modules/ui/animation/constants';
 import { useConfigContext } from '@/modules/config/hooks/useConfigContext';
 import { QueryParams, mapIntentToQueryParam } from '@/lib/constants';
-import { requiresMainnet } from '@/lib/widget-network-map';
+import { requiresMainnet, isMultichain } from '@/lib/widget-network-map';
 import { isL2ChainId } from '@jetstreamgg/sky-utils';
 import { normalizeUrlParam } from '@/lib/helpers/string/normalizeUrlParam';
 import { LinkedActionWrapper } from '@/modules/ui/components/LinkedActionWrapper';
@@ -22,6 +22,7 @@ import { Sheet, SheetContent, SheetTrigger } from '@/components/ui/sheet';
 import { Button } from '@/components/ui/button';
 import { DualSwitcher } from '@/components/DualSwitcher';
 import { useNetworkSwitch } from '@/modules/ui/context/NetworkSwitchContext';
+import { useChains } from 'wagmi';
 import {
   Tooltip,
   TooltipContent,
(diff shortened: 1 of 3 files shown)


iceanddust (approved):
Nice! It works really well, I followed the testing steps and couldn't find any issues.

A couple of thoughts/comments:
- I couldn't reproduce the issue that codewiththeflowxyz mentioned, maybe that was the expected behavior as mainnet was the currently saved chain for savings? Or maybe a race condition.
- While testing the race condition, I noticed that if you switch fast enough to trick the app into trying to render a widget that is not available for the destination chain, it redirects to the balances widget. Maybe this is fine though as it's an edge case handling, and it doesn't result in a blank widget.
- It feels a bit odd to me that both multi-chain widgets track different preferences, perhaps they should handle the same preference? I think in practice, given how similar both widgets are and that they handle similar tokens, it would be less likely that the user wants to perform different actions for the 2 widgets, I see both the L2 savings and trade widgets kind of linked to each other.
- This is not introduced with this feature, but it made it more evident to me, however we could further discuss it later: I think for a newbie user it's not that straight-forward to understand which chains are available for each module (given that not all modules are available for all supported chains). Maybe we could improve the comms here by showing a list of the chains supported for each module? I had thought of something like the following image, but perhaps it adds too much noise to the widgets: 

Image

PR #768, opened 2025-09-05
Ba labs data for rate chart, stusds suppliers, expert chart

inclinedmind:
- renames useParseSavingsChartData to useParseTvlChartData
  - Adds rate chart functionality to stUSDS with dual TVL/rate chart tabs
  - Implements stUSDS supplier count display using real data from useOverallSkyData
  - Switches expert chart to use dedicated useStUsdsChartInfo hook instead of generic token chart hook
  - Updates chart offset calculation to handle percentage values properly
  - Removes unnecessary margin in the upgrade stats card

  Testing steps:

  1. Navigate to stUSDS page and verify both TVL and Rate chart tabs display correctly
  2. Check that stUSDS supplier count shows actual data (0) instead of "TODO"
  3. Verify expert chart displays stUSDS TVL data correctly
  4. Test savings, seal, and stake charts still function properly after hook refactor
  5. Confirm chart percentage calculations now works correctly for rate charts


bitsforpeace (approved):
The Expert and stUSDS charts all show 0 for me. I assume it's because BA labs doesn't support tenderly?

There were a couple minor coderabbit comments that would be good to include.

PR #793, opened 2025-09-12
Hide expert modules and stUSDS UI elements behind a feature flag

iceanddust:
### What does this PR do?

This PR introduces a new feature flag, `VITE_ENABLE_EXPERT_MODULES`, to conditionally enable or disable all features related to "Expert Modules".

When the flag is disabled, this change will:
*   Hide the "Expert" tab from the main widget navigation.
*   Remove `stUSDS` from all balance lists and portfolio views.
*   Prevent any "suggested actions" that lead to expert modules (e.g., upgrading to `stUSDS`).
*   Block direct URL access to expert modules by invalidating the relevant URL parameters.

### Testing steps:

**1. When the feature flag is enabled (`VITE_ENABLE_EXPERT_MODULES=true`):**
    *   Run the application.
    *   Verify the app behaves as it currently does on `development`.
    *   Confirm the "Expert" tab is visible and functional.
    *   Confirm `stUSDS` is visible in balance lists and that `stUSDS`-related suggested actions appear.

**2. When the feature flag is disabled (variable is `false` or not set):**
    *   Set `VITE_ENABLE_EXPERT_MODULES=false` in your `.env` file and restart the app.
    *   Verify the "Expert" tab is hidden from the widget navigation.
    *   Check the Portfolio and Balances views to confirm `stUSDS` does not appear.
    *   Verify that no suggested actions related to `stUSDS` are displayed.
    *   Attempt to navigate directly to an expert module via URL (e.g., `/?widget=expert&expert_module=stusds`). Confirm that the URL parameters are removed and the user is not taken to the expert widget.


apps/webapp/src/modules/app/components/WidgetPane.tsx
```diff
@@ -101,9 +102,23 @@ export const WidgetPane = ({ intent, children }: WidgetPaneProps) => {
...
+  // Filter out stUSDS from token list if expert modules are disabled
+  const filteredBalancesTokenList = isExpertModulesEnabled()
+    ? defaultConfig.balancesTokenList
+    : Object.entries(defaultConfig.balancesTokenList).reduce(
+        (acc, [chainId, tokens]) => {
+          acc[Number(chainId)] = tokens.filter(token => token.symbol !== 'stUSDS');
+          return acc;
+        },
+        {} as typeof defaultConfig.balancesTokenList
+      );
```
bitsforpeace, line 121: Would it make more sense to configure this in the `tokenListConfig` file? Instead of running the filtering logic everytime `WidgetPane` renders?

Regardless, I think we would want to import the token info from `tokens.constants.ts` and use something like `stusds.symbol` rather than the string value.

iceanddust:
I've implemented the suggestions above. The unrestricted balances token list filtering couldn't be moved to where it is defined because it currently lives in the widgets package, and we can't access env variables there, so I moved it to the `defaultConfig` definition instead.
Other than that, I noticed that neither sUSDS nor stUSDS is defined in any of the restricted token lists (restricted and restrictd MiCA). Should it be added to any of those?

PR #804, opened 2025-09-17
Update chatbot confirmation warning UI

codewiththeflowxyz:
### How to test it

Ask "How can I upgrade 10 DAI" and click the suggested action

https://github.com/user-attachments/assets/166cceed-b294-4256-bc13-9c40c17b8b4a

### DIFF
--- a/apps/webapp/.content-version
+++ b/apps/webapp/.content-version
@@ -1,1 +1,1 @@
-main
+bfp/speed-bump-copy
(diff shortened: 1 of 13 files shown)


generatorpoint:
> I see the speedbump when I click "Upgrade 10 DAI to USDS" but not when I click "Go to Upgrade" - is that expected?

I believe so because I think the speed bump is for when the inputs become prefilled due to the intent, however we should verify with @codewiththeflowxyz

PR #805, opened 2025-09-17
add speed-bump extraction from corpus

generatorpoint:


inclinedmind (approved):

codewiththeflowxyz (approved):
Tested it locally and it works great. We need to replace Seal with the Staking engine, but this can be done later in a separete PR.

PR #822, opened 2025-09-22
Update speed bumps with expert and staking modules

codewiththeflowxyz:
Adds speed-bump content for the staking engine and the expert module

### DIFF
--- a/packages/widgets/src/data/tooltips/index.ts
+++ b/packages/widgets/src/data/tooltips/index.ts
@@ -21,7 +21,7 @@ Open Safe[Wallet]`
   },
   {
     id: 'borrow-rate',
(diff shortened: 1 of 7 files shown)


packages/widgets/src/data/tooltips/index.ts
```diff
@@ -21,7 +21,7 @@ Open Safe[Wallet]`
   },
   {
     id: 'borrow-rate',
-    title: 'Borrow Rate',
+    title: 'Borrow rate',
```
bitsforpeace, line 24: Is this coming from the corpus repo? It was recently pointed out that it should be captalized, "Borrow Rate"?

bitsforpeace (approved):

PR #832, opened 2025-09-22
Fix chain preference when navigating between 2 multichain widgets

iceanddust:
### What does this PR do?

This PR enhances the network auto-switching logic to restore a user's last-used network when switching between different multichain widgets.

Previously, the app would only restore a saved network preference if the user was coming *from* a mainnet-only widget. This change adds a second case, ensuring the network preference is also restored when navigating directly from one multichain widget (e.g., Trade) to another (e.g., Savings) directly.

### Testing steps:

1.  Navigate to a multichain widget (e.g., Trade) and select an L2 network (e.g., Base).
2.  Navigate to a different multichain widget (e.g., Savings) and select a different L2 network (e.g., Arbitrum).
3.  Switch back to the Trade widget.
4.  **Verify:** The wallet network should automatically switch back to Base.
5.  Switch back to the Savings widget.
6.  **Verify:** The wallet network should automatically switch back to Arbitrum.

### DIFF
--- a/apps/webapp/src/modules/ui/components/ChainModal.tsx
+++ b/apps/webapp/src/modules/ui/components/ChainModal.tsx
@@ -96,7 +96,7 @@ export function ChainModal({
         onCloseAutoFocus={e => e.preventDefault()}
       >
         <DialogTitle>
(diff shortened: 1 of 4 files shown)


inclinedmind (approved):
Nice 👍  fyi there's a console log that can be removed

PR #841, opened 2025-09-23
Multichain UX - Return to origin

bitsforpeace:
### What does this PR do?

Updates multichain UX switch logic:

1. **Manual network changes always win**
    
    - If the user explicitly selects a network (via app selector or wallet), switch to that network and keep using it.
        
2. **Entering a mainnet-only widget**
    
    - If the current network is **not** Mainnet, save the **origin** as the current network, then force switch to Mainnet.
        
    - If already on Mainnet, do nothing (do not set or change origin).
        
3. **Navigating from mainnet-only to a multichain widget**
    
    - If an **origin** exists, switch to the origin network **once**, then **clear the origin**.
        
    - If no origin exists, stay on the current network (Mainnet).
        
4. **Navigating between multichain widgets**
    
    - Do not change networks automatically. Continue using the current network.
        
5. **Re-entering mainnet-only while on Mainnet**
    
    - No change to network or origin.
        
6. **Repeated mainnet-only visits without leaving mainnet-only**
    
    - No change to network or origin. The origin remains whatever it last was until you eventually leave mainnet-only for a multichain widget.
        
7. **Manual network changes while in mainnet-only**
    
    - If the user manually switches networks while in a mainnet-only widget, and the widget truly requires Mainnet, the app redirects to the Balances widget.
        
8. **Clearing the origin**
    
    - The origin must be cleared **immediately after** it is used to return from Mainnet to a multichain widget, so it does not affect subsequent navigations.

### DIFF
--- a/apps/webapp/src/modules/app/components/WidgetMenuItemTooltip.tsx
+++ b/apps/webapp/src/modules/app/components/WidgetMenuItemTooltip.tsx
@@ -15,7 +15,6 @@ interface WidgetMenuItemTooltipProps {
   description?: string;
   widgetIntent: Intent;
   currentChainId?: number;
-  currentIntent?: Intent;
   label: string;
   isMobile: boolean;
   disabled?: boolean;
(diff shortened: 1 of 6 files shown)


generatorpoint:
closing this because we went with a different network switching strategy

PR #846, opened 2025-09-23
e2e test for network switching behavior

generatorpoint:


generatorpoint:
I think @0x2tester has a better approach to this, we can close this branch

PR #847, opened 2025-09-23
Fix padding alignment issues across the widgets

iceanddust:
### What does this PR do?
Ensures that all widgets have consistent paddings in both left and right sites in all breakpoints

### Testing steps:
- Navigate to different widgets in different breakpoints
- Verify that they are balanced on both sides: they have the same space on the left and right, and all elements in the widget pane are properly aligned on both sides


bitsforpeace (commented):
I changed the target branch to development. I think we can aim to merge all that code together for the next release. 

I tested the PR, but the padding on the Connect button on the Balances widget looks lower than all the other widgets for me

generatorpoint (approved):

PR #855, opened 2025-09-25
Replace toast with sonner

codewiththeflowxyz:
Note for testing: The easiest way to trigger multiple notifications is by changing networks multiple times

https://github.com/user-attachments/assets/d6c859ab-58bd-49a0-9f5d-263d8b62a649


pnpm-workspace.yaml
```diff
@@ -67,6 +67,7 @@ catalog:
   knip: ^5.63.0
   lint-staged: ^15.5.1
   lucide-react: ^0.544.0
+  next-themes: ^0.4.6
```
iceanddust, line 70: Do we need this dependency? I see ShadCN uses it in the sonner component in their example, but we don't seem to be using it

iceanddust (commented):
I like where this is going, and the stacked cards look good! I did find a couple of edge cases worth taking a look to:

1. The 'close all' button is overlapping the footer links. Maybe we can move the toast container up a bit, or move the footer links to the bottom-left corner
image

2. The gap between each toast looks like is not considered part of the container, so there's a weird jumping animation when hovering the mouse there:

https://github.com/user-attachments/assets/e420cf8c-0f50-4ab9-a423-19fd68126b78

3. On mobile, the toasts need to be moved up a bit because the first toast is not fully visible

https://github.com/user-attachments/assets/f3af1c9d-e015-44c0-bad5-63bec7cd5745

codewiththeflowxyz:
> I like where this is going, and the stacked cards look good! I did find a couple of edge cases worth taking a look to:
> 

1. I'll move the footer links to the left
2. Good catch, I'll take a look
3. I can't replicate it

iceanddust (approved):

bitsforpeace (approved):

PR #856, opened 2025-09-25
Close chatbot after clicking on a suggested action when on mobiles

codewiththeflowxyz:
On mobiles the chatbot covers the whole screen, after following a suggested action we need to dismiss the chatbot screen.


iceanddust (approved):

inclinedmind (approved):
lgtm. coderabbit left some comments but not sure if they make sense and I don't see any issues.

PR #869, opened 2025-09-30
Set step to 2 on flows that don't require allowance

iceanddust:
### What does this PR do?
It fixes an issue where some flows that don't require allowance would not render the correct color and icon to the card on a successful transaction, given that it was detecting an incorrect intermediate step.
Fixes this issue https://app.shortcut.com/dux-makerdao/story/3800/fix-success-icon-in-rewards-flow

### Testing steps:
- Attempt to perform a rewards claim, savings withdrawal, or staking engine claim.
- Upon a successful transaction, the card should be colored purple, and the check icon should be used


inclinedmind (approved):

codewiththeflowxyz (approved):
I've tested it and it works as expected.

bitsforpeace (approved):

PR #871, opened 2025-10-01
Adjust about cards height dynamically

codewiththeflowxyz:
https://github.com/user-attachments/assets/93290eb5-9461-4de2-b109-a14287712f31


inclinedmind (approved):
Works well for me 👍

bitsforpeace (approved):

PR #1389, opened 2026-03-13
Add consent-aware person_profiles to PostHog config

codewiththeflowxyz:
## Summary
  - Adds `person_profiles` to `posthog.init()` — set to `'always'` for accepted users and `'identified_only'` for pending-consent users
  - On consent granted, upgrades to `'always'` via `set_config()`; on consent revoked, downgrades back to `'identified_only'`
  - Without this setting, the SDK defaults to `identified_only` which requires `posthog.identify()` calls — resulting in "No defined properties" for all persons in insights

Related Linear ticket: APP-119

  ## Test plan
  - [ ] Run locally, open DevTools Console (PostHog debug mode logs events)
  - [ ] Before accepting cookies: verify events include `$process_person_profile: false`
  - [ ] After accepting cookies: verify events include `$process_person_profile: true`
  - [ ] Check PostHog dev project — consented persons should show browser/OS properties


inclinedmind (approved):
tested it out locally and saw the process_person_profile switch from false to true after accepting

bitsforpeace (approved):

PR #1403, opened 2026-03-18
Implement geo-based cookie consent logic in CookieConsentBanner and u…

codewiththeflowxyz:
…pdate GeoConfig context. The banner now auto-accepts consent if not required for the user's geography. Adjusted visibility conditions and added new properties to GeoConfig for cookie banner management.


iceanddust (approved):
LGTM, the only thing I noticed is that when cookie banner is not required, you still see the footer link for 'Cookie Settings', but clicking it does nothing. I think we should either hide it when `isCookieBannerRequired` is false, or make the banner be able to be opened only when clicking the link

PR #1471, opened 2026-04-09
feat: pair-aware auto slippage for cowswap trades

iceanddust:
## Summary
- Use **0.05%** auto slippage for stablecoin pairs (USDC/USDT/DAI/USDS) and **0.3%** for volatile pairs, replacing the flat 0.5% default
- ETH flow slippage unchanged (2% mainnet, 0.5% L2)
- Users can still override with Custom mode

Closes APP-130

## Test plan
- [x] Select USDC→USDS: auto mode shows 0.05%
- [x] Select USDS→USDT: auto mode shows 0.05%
- [x] Select SKY→USDS: auto mode shows 0.3%
- [x] Select ETH→USDS: auto mode shows 2% (ETH flow unchanged)
- [x] Switch from Auto to Custom: input pre-fills with pair-aware default
- [x] Switch tokens while in auto mode: displayed slippage updates
- [x] Execute a stablecoin swap and verify correct slippage in quote


codewiththeflowxyz:
@codex

packages/widgets/src/widgets/TradeWidget/components/TradeConfigMenu.tsx
```diff
@@ -88,7 +94,7 @@ export const TradeConfigMenu = ({
               className="w-full"
               defaultValue={slippage ? SlippageType.CUSTOM : SlippageType.AUTO}
               onValueChange={value => {
-                const newSlippageValue = value === SlippageType.AUTO ? '' : slippageConfig.default.toString();
+                const newSlippageValue = value === SlippageType.AUTO ? '' : autoSlippage.toString();
```
codewiththeflowxyz: The naming here is a bit confusing, because if SlippageType is different that AUTO we set it as `autoSlippage`. Maybe a renaming of the `autoSlippage` const or a comment would do.

codewiththeflowxyz (approved):
LGTM, I left a comment about a confusing line though.

PR #1492, opened 2026-04-15
Add regression test for correct sUSDS preview values in transaction overview

codewiththeflowxyz:


inclinedmind (approved):
Good call. I think we could read the susds amount moved on chain and compare against that, but we can make that improvement in another PR

bitsforpeace (approved):

PR #1509, opened 2026-04-22
pnpm.overrides: prune redundant entry, tighten replacements, document conventions

bitsforpeace:
## Summary

Audit-driven cleanup of `pnpm.overrides` as a follow-up to [Wave 1](https://github.com/jetstreamgg/tarmac/pull/1506) / [Wave 2](https://github.com/jetstreamgg/tarmac/pull/1508). Part of [APP-198](https://linear.app/skybasestar/issue/APP-198).

### Changes

1. **Remove the `vite@>=7.0.0 <7.3.2 → ^7.3.2` override.**  Redundant. The catalog bump to `vite: ^7.3.2` in `pnpm-workspace.yaml` (Wave 1) covers every direct consumer, and `pnpm why -r vite` reports **only** 7.3.2 in the tree — no transitive ever lands in the `<7.3.2` range, so the override is doing nothing.

2. **Tighten Wave 1 replacement values to tilde ranges** for `protobufjs`, `dompurify`, `lodash`, `defu`, `h3`. Unbounded `>=X.Y.Z` replacements let pnpm's deduplication silently lift older-major consumers onto a newer major — the exact failure I hit on picomatch / bn.js / glob during Wave 2, where a `^2.2.1` spec was resolving to `picomatch 4.x`. Today each of these Wave 1 packages has only a single major in the tree, so this is defensive — it stops the failure mode from re-emerging if a future transitive drags in an older major.

3. **Add a `//pnpm.overrides` comment block** above the overrides documenting the two conventions that future additions should follow:
   - Range-scoped selector (`pkg@<patched`) so the override self-expires once upstreams publish clean versions.
   - Tilde-or-tighter replacement value to stay inside the same minor.

### Left in place (deliberate)

- `esbuild@<=0.24.2 → ^0.25.0` and `@babel/runtime@7.26.7 → 7.26.10` — both pre-existing, both currently inert (no matching version in the tree). Zero runtime cost and we don't have the original context. Leaving alone; if someone turns up the reason they can be cleanly removed later.
- `axios@>=1.0.0 <1.15.0 → 1.15.0` and `@json-rpc-tools/provider>axios@^0.21.0 → 1.15.0` — both active security pins for ongoing axios CVEs. The targeted one is the only thing keeping `@json-rpc-tools/provider 1.7.6` (pulled in via `eip1193-provider`) off axios 0.21.x.

### Resolved versions after cleanup (unchanged)

```
protobufjs 7.5.5
dompurify  3.4.0
lodash     4.18.1
defu       6.1.7
h3         1.15.11
vite       7.3.2   (only version in tree)
```

## Test plan

- [x] `pnpm install` succeeds with the cleaned overrides
- [x] `pnpm typecheck` passes across all workspaces
- [x] `pnpm why -r` confirms every previously-patched package is still at its patched version
- [ ] CI green

## Follow-ups

- **Wave 2 PR ([#1508](https://github.com/jetstreamgg/tarmac/pull/1508))** still has some `>=X.Y.Z` replacements that should be tightened to tilde for consistency (`flatted`, `follow-redirects`, `js-yaml`, `qs`, `hono`, `ajv`, `mdast-util-to-hast`, and the minimatch / glob / rollup entries). I'll follow up with either a commit on that branch before merge or a second cleanup PR afterwards — same idea, same convention.
- **Track down the history of `esbuild@<=0.24.2` and `@babel/runtime@7.26.7`** — if the reason they were added no longer applies, drop them in a later cleanup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,18 +104,22 @@
     "typescript": "catalog:",
     "vitest": "catalog:"
   },
+  "//pnpm.overrides": [
+    "Conventions for entries below:",
+    "1. Prefer range-scoped selectors (pkg@<patched) so the override self-expires once upstreams publish clean versions.",
+    "2. On the replacement side, use a tilde range (~X.Y.Z) pinned to the same minor whenever multiple majors of the package may coexist in the tree - an unbounded >=X.Y.Z can silently bump older-major consumers to a newer major via pnpm's dedup (see picomatch/bn.js/glob)."
+  ],
   "pnpm": {
     "overrides": {
       "esbuild@<=0.24.2": "^0.25.0",
       "@babel/runtime@7.26.7": "7.26.10",
       "axios@>=1.0.0 <1.15.0": "1.15.0",
       "@json-rpc-tools/provider>axios@^0.21.0": "1.15.0",
-      "protobufjs@<7.5.5": "^7.5.5",
-      "dompurify@<3.4.0": ">=3.4.0",
-      "lodash@<4.18.0": ">=4.18.0",
-      "defu@<6.1.5": ">=6.1.5",
-      "h3@<1.15.9": ">=1.15.9",
-      "vite@>=7.0.0 <7.3.2": "^7.3.2"
+      "protobufjs@<7.5.5": "~7.5.5",
+      "dompurify@<3.4.0": "~3.4.0",
+      "lodash@<4.18.0": "~4.18.0",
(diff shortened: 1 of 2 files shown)


bitsforpeace:
Superseded by #1508 — the same tightening (plus conventions comment and vite-override drop) is now applied directly on the Wave 2 branch. No need to ship this one separately.

PR #1516, opened 2026-04-24
Bump vitest + @vitest/coverage-v8 to v4 (APP-212)

bitsforpeace:
- Catalog: vitest ^3.2.4 → ^4.0.0, @vitest/coverage-v8 ^3.2.4 → ^4.0.0, vitest-canvas-mock ^0.3.3 → ^1.1.4 (v4 peer compat)
- packages/hooks: migrate removed singleFork poolOption to maxWorkers:1 / isolate:false
- apps/webapp vite config: exclude e2e specs that v4 now picks up
- Widget tests: ResizeObserver mock uses function() {} so v4 strict vi.fn() allows `new` construction
- Webapp tests with partial module mocks: migrate to importOriginal pattern (v4 no longer tolerates accessing non-mocked exports)
- PsmConversionDetails test: fix stale mock path and title text (was mocking AboutUsds but component imports AboutPsmConversion)

Known: ConvertWidgetPane and BalancesSuggestedActions each have one semantic assertion failure under v4 that needs further triage — text matchers can't find expected buttons after importOriginal spreads pull in real modules. Tracking separately.

### DIFF
--- a/apps/webapp/src/modules/convert/components/PsmConversionWidgetPane.test.tsx
+++ b/apps/webapp/src/modules/convert/components/PsmConversionWidgetPane.test.tsx
@@ -22,12 +22,16 @@ const onAnalyticsEventMock = vi.fn();
 
 let capturedWidgetProps: Record<string, any> | undefined;
 
-vi.mock('@jetstreamgg/sky-widgets', () => ({
-  PsmConversionWidget: (props: Record<string, any>) => {
-    capturedWidgetProps = props;
-    return <div data-testid="psm-conversion-widget" />;
-  }
-}));
+vi.mock('@jetstreamgg/sky-widgets', async importOriginal => {
+  const actual = await importOriginal<typeof import('@jetstreamgg/sky-widgets')>();
+  return {
+    ...actual,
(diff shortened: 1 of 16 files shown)


apps/webapp/src/modules/convert/components/PsmConversionWidgetPane.test.tsx
```diff
@@ -43,9 +47,13 @@ vi.mock('@/modules/analytics/hooks/useWidgetAnalytics', () => ({
...
-}));
+vi.mock('wagmi', async importOriginal => {
+  const actual = await importOriginal<typeof import('wagmi')>();
+  return {
+    ...actual,
+    useChainId: () => 1
+  };
+});
 
 vi.mock('react-router-dom', () => ({
```
iceanddust, line 58: Does this needs to be updated too, the same way it was updated in https://github.com/jetstreamgg/tarmac/pull/1516/changes#diff-f5f283dd463c53639408bac155eb16dabd4fac06c3aaf0495e3e0c7797148942R65 ?

iceanddust (approved):
LGTM, just a small comment

PR #1524, opened 2026-04-28
Bump recharts 2 → 3 (APP-217)

codewiththeflowxyz:
## Summary
- Bumps `recharts` from `^2.15.4` to `^3.8.1` via the `pnpm-workspace.yaml` catalog.
- Zero source changes required — our chart consumption (one file: `apps/webapp/src/modules/ui/components/Chart.tsx`) uses only foundational primitives (`<AreaChart>`, `<Area>`, `<XAxis>`, `<YAxis>`, `<Tooltip>`, `<ResponsiveContainer>`) with stable props (`dataKey`, `stroke`, `type`, `domain`, `padding`, `axisLine`, `tickLine`).

## Why it's safe
- **One direct importer.** `Chart.tsx` is the only file in the repo that imports from `recharts`. Every other chart in the app (savings, stake, morpho vault, upgrade, rewards) is a wrapper around this single component. The ticket's expectation of widget-package consumers turned out to be stale; those charts no longer exist in `packages/widgets`.
- **Hand-rolled tooltip type.** `ChartTooltip.tsx` declares its own `CustomTooltipProps` interface rather than importing recharts' `TooltipProps`, which insulated us from v3's type tightening.
- **CSR-only.** v3's SSR rework doesn't apply to the webapp.
- **Lockfile diff (280 lines) ≫ source diff (0 lines).** Recharts v3 reshuffled its transitive dep tree significantly without breaking the public API surface our consumers touch.

## Validation
- ✅ `pnpm install` clean (only pre-existing peer warnings unrelated to this bump).
- ✅ `pnpm typecheck` clean across all four workspaces.
- ✅ `pnpm -F @jetstreamgg/sky-utils test` — 52/52 pass.
- ✅ `pnpm build` (production) succeeds.

## Test plan
- [ ] **Tooltip on hover** — savings, stake, morpho vault, upgrade, rewards charts. Tooltip should show date label, value, and Min/Max badge where applicable. (Highest-risk visual surface — v3 normalized the `payload` shape.)
- [ ] **Min/Max dots** render at the lowest and highest points of each chart line.
- [ ] **Responsive resizing** — chart fits its card at desktop and mobile breakpoints; resize the window and confirm no overflow or collapse.
- [ ] **Timeframe controls** (1W/1M/1Y/All) re-render the chart and the external x-axis label strip.
- [ ] **Empty/loading states** still show the skeleton; error state still shows the warning UI.
- [ ] No `recharts`-related console warnings or errors while navigating.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -250,8 +250,8 @@ catalogs:
       specifier: ^6.30.3
       version: 6.30.3
     recharts:
-      specifier: ^2.15.4
-      version: 2.15.4
+      specifier: ^3.8.1
+      version: 3.8.1
     rehype-sanitize:
       specifier: ^6.0.0
       version: 6.0.0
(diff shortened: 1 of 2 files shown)


iceanddust (approved):
LGTM, I just noticed 1 small weird behavior, but I think that's outside the scope of this PR. We could create a Linear issue for this and merge this one.

This is the issue: when I hover over the chart, the tooltip always renders in the top-left corner and moves to the pointer position. I can see this behavior also in prod, but only sometimes

This branch:

https://github.com/user-attachments/assets/55b6d4ec-0db4-4196-b456-835aab96acfb

Production:

https://github.com/user-attachments/assets/509c49ae-30df-4159-b8a7-cdc8dc7888ed

PR #1530, opened 2026-04-29
Pendle hooks layer (APP-175)

iceanddust:
## Summary

Hooks layer for the Pendle Fixed Yield integration ([APP-175](https://linear.app/skybasestar/issue/APP-175/pendle-integration)).

Adds `packages/hooks/src/pendle/` with everything needed for Buy / Withdraw / Exit flows on the PT-USDG market.

- **API client** (`pendleApiClient.ts`) — `fetchPendleConvert` (POST `/v3/sdk/1/convert`) and `fetchPendleMarketsByIds` (GET `/v2/markets/all`). Hardcoded to `mainnet.id` since our integration only adds mainnet markets and Pendle's API doesn't serve Tenderly anyway (Tenderly fork submits the resulting tx against forked mainnet state).
- **Read hooks** — `useQuotePendleConvert` (per-trade quote, 60s refetch), `usePendleMarketsApiData` (one HTTP call returns headline stats for every market in `PENDLE_MARKETS`, keyed for both overview-list and detail-view consumers), `usePendleUserPtBalances` (batched user PT balances across all markets).
- **Batch-aware write hook** — `useBatchPendleConvert` uses the existing `useTransactionFlow` primitive to bundle approval + convert into a single signature on EIP-5792 wallets, with sequential fallback otherwise. Mirrors the `useBatchMorphoVaultDeposit` pattern.
- **Security primitive** (`buildVerifiedArgs.ts`) — the only thing standing between a compromised `/convert` response and a user signature. See "Security model" below.
- **Config** (`constants.ts`) — `PENDLE_MARKETS` array (PT-USDG seeded), pinned router address, minimal Router ABI containing only the three allowlisted selectors, empty-struct constants matching Pendle's [`StructGen.sol`](https://github.com/pendle-finance/pendle-examples-public/blob/main/src/StructGen.sol).
- **Helpers** — `isMarketMatured`, `secondsToExpiry`.

## Security model

Pendle's `/convert` endpoint returns ready-to-execute calldata. Trusting it blindly would mean a compromised API could redirect funds, route through an attacker-controlled aggregator (`pendleSwap` + `swapData.extCalldata`), or fill malicious limit orders (`limit.optData`). Wallets do **not** decode these nested struct fields meaningfully, so we cannot rely on the user's wallet UI to catch tampering.

`buildVerifiedArgs` defends against this by rebuilding the call from values **we** control and forcing every dangerous field to its empty form before viem encodes calldata for signing:

- **Pinned router** — `useBatchPendleConvert` always submits to `PENDLE_ROUTER_V4_ADDRESS` from our config, never to anything from the API.
- **Selector allowlist** — `swapExactTokenForPt` (Buy), `swapExactPtForToken` (Withdraw pre-maturity), `exitPostExpToToken` (Withdraw post-maturity). Anything else (including `redeemPyToToken`, limit-order paths, liquidity ops) → reject.
- **Override** — `receiver`, `market`, `tokenIn`/`tokenOut`, `netTokenIn`/`exactPtIn`/`netPtIn`, and `minOut` (passed through from API; if compromised, the other guards still bound the blast radius).
- **Force empty** — `pendleSwap = 0x0`, `swapData = empty`, `limit = empty`, `tokenMintSy === tokenIn` / `tokenRedeemSy === tokenOut` (no-aggregator invariant), `netLpIn = 0n` for exit (v1 has no LP). The result is byte-equivalent to Pendle's `createTokenInputStruct` / `emptyLimit`.

## Selector matrix

| Flow | Method | Args | Notes |
|---|---|---|---|
| Buy | `swapExactTokenForPt` | `(receiver, market, minPtOut, guessPtOut, input, limit)` | `guessPtOut` solver hint passed through to save ~180k gas vs `defaultApprox` |
| Withdraw, pre-maturity | `swapExactPtForToken` | `(receiver, market, exactPtIn, output, limit)` | |
| Withdraw, post-maturity | `exitPostExpToToken` | `(receiver, market, netPtIn, netLpIn, output)` | Triggered by passing YT in API request `inputs[1]` with amount 0; widget passes `ytTokenForExit` iff matured |

## Test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` clean
- [x] 19/19 unit tests pass — covers all three security branches:
  - **Buy**: tampered `receiver` / `market` / `netTokenIn` / `tokenMintSy` overridden; `pendleSwap` + `swapData` + `limit` force-empty; `apiMinOut` passed through; selector allowlist rejects unknown / wrong-flow methods; malformed `guessPtOut` rejected
  - **Withdraw (Sell)**: `tokenRedeemSy === tokenOut`, no-aggregator invariant, selector allowlist
  - **Exit (post-maturity)**: `netLpIn` force-empty, `receiver` override, `tokenRedeemSy` invariant
- [ ] Integration / E2E tests come in a later PR (widget + webapp module work)

## Notes

- Skipped wagmi codegen for the Pendle Router — it's a multi-action proxy whose Etherscan `getabi` returns only the empty proxy ABI. Hand-wrote a minimal ABI inline with only the three allowlisted selectors. This doubles as a structural allowlist: any non-allowlisted selector throws at the viem encoding layer too, on top of our explicit `PENDLE_ALLOWED_SELECTORS` check.
- The matured-market path was confirmed against a real `/convert` response (not speculation): pre-maturity `swapExactPtForToken` and post-maturity `exitPostExpToToken` are distinct functions with different signatures — `exitPostExpToToken` has no `limit` parameter and includes `netLpIn` (forced to 0n in v1).
- This PR targets the long-lived `feature/app-175-pendle-integration` branch. Subsequent PRs (widgets, webapp module, balances integration, E2E) will stack onto it.


codewiththeflowxyz (approved):
LGTM! Claude found one issue though.

codewiththeflowxyz:
### Code review

Found 1 issue:

1. `extractApiMinOut` silently returns `0n` when the API response is missing or non-numeric, and that value flows verbatim into the on-chain `minPtOut`/`minTokenOut`. A malformed `/convert` response — or a missing `params[2]` / `output.minTokenOut` field on any of the three allowlisted methods — yields a signed transaction with zero slippage protection (full sandwich exposure), bypassing the security model the rest of `buildVerifiedArgs` defends. Either throw from `extractApiMinOut` (so react-query surfaces it as a quote error) or compute a local floor from `quote.amountOut * (1 - slippage)` — the `slippage` parameter is already required by `KnownCallValues` but currently unused.

https://github.com/jetstreamgg/tarmac/blob/ba303b39d410ac8bf41b455f4546899415ea90a7/packages/hooks/src/pendle/useQuotePendleConvert.ts#L9-L34

https://github.com/jetstreamgg/tarmac/blob/ba303b39d410ac8bf41b455f4546899415ea90a7/packages/hooks/src/pendle/buildVerifiedArgs.ts#L204-L257

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/hooks/src/pendle/useQuotePendleConvert.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (method === 'exitPostExpToToken') {
+      // (receiver, market, netPtIn, netLpIn, output)
+      const output = params[4] as { minTokenOut?: string } | undefined;
+      return BigInt(output?.minTokenOut ?? 0);
+    }
+    return 0n;
+  } catch {
+    return 0n;
+  }
+}
```
codewiththeflowxyz, line 41: **Silent `0n` fallback removes slippage protection.**

When `params[2]` (Buy) or `output.minTokenOut` (Withdraw / ExitPostExp) is missing or non-numeric, this returns `0n` instead of failing. That `0n` then flows verbatim into the on-chain `minPtOut` / `minTokenOut` via `quote.apiMinOut` in [`buildVerifiedArgs.ts#L204-L257`](https://github.com/jetstreamgg/tarmac/blob/ba303b39d410ac8bf41b455f4546899415ea90a7/packages/hooks/src/pendle/buildVerifiedArgs.ts#L204-L257) — i.e. one malformed `/convert` response = signed transaction with zero slippage protection (full sandwich exposure), bypassing the rest of the override matrix.

The `slippage` parameter is already required by `KnownCallValues` and threaded through `useBatchPendleConvert` but never read inside `buildVerifiedArgs`. Two reasonable fixes:

1. Throw from `extractApiMinOut` on missing/invalid input so react-query surfaces a quote error and the user sees "couldn't get a quote" instead of signing.
2. Use the already-passed `slippage` to compute a local floor (`quote.amountOut * (1 - slippage)`) and clamp `apiMinOut` against it inside `buildVerifiedArgs` — also removes the dead `slippage` field.

The JSDoc on `amountOut` in `pendle.d.ts` and on this file actually claim option (2) is what the code does today — doc and impl are out of sync, which is a separate signal that this is the intended behavior.

codewiththeflowxyz (commented):
Inline note for the issue raised above.

PR #1551, opened 2026-05-08
Clean up production console errors and warnings APP-195

inclinedmind:
Fixes

  - GA / GTM blocked by CSP — added Google Tag Manager / Analytics domains to script-src, img-src, connect-src per Google's CSP guide. GA loader script can now actually run.
  - Inline-script CSP block from getProvider — passed preference: { telemetry: false } to the Base Account connector. Stops @base-org/account from injecting an inline telemetry script that our CSP blocks.
  - Radix Missing Description dialog warnings — added aria-describedby={undefined} to dialogs without a <DialogDescription>
  - Radix Missing DialogTitle warning on the mobile widget nav — WidgetNavigation's <SheetContent> rendered the "Menu" heading as a plain <Heading>, so Radix's accessibility check fired (Sheet is built on @radix-ui/react-dialog). Wrapped it in <SheetTitle asChild> and added aria-describedby={undefined} to match the suppression pattern used on the Dialogs.
  - Unknown tooltip type: sky-savings-rate / stusds-rate — added a resolvePopoverTooltipKey helper in widgets that reverse-resolves a full id back to its short key. Banner parser and FAQ accordion now both render through PopoverRateInfo regardless of which form authors write.
  - Repeated safe-transaction-*.safe.global 404s — useIsSafeWallet now caches the result for the session (staleTime: Infinity, gcTime: Infinity), skips the call entirely when connector.id === 'safe', and the isSafeConnector flag is gated on address so it can't flip true mid-connecting.

  Test plan

  - Open /savings and /stusds — confirm no Unknown tooltip type warning, tooltip popovers open with correct content
  - Open Connect / Chain / Connected / Transaction modals — confirm no Radix dialog warning
  - Open the mobile widget nav (hamburger) — confirm no Radix DialogTitle / Description warning
  - Connect via Base Account — confirm no inline-script CSP error
  - Connect a non-Safe wallet — confirm exactly one /api/v1/safes/{address} 404 per session, not one per modal open / focus
  - GA fix can only be tested on prod, since its only used on prod

### DIFF
--- a/apps/webapp/src/data/wagmi/config/config.default.ts
+++ b/apps/webapp/src/data/wagmi/config/config.default.ts
@@ -37,7 +37,8 @@ const connectors = [
   metaMask(),
   baseAccount({
     appName: 'sky.money',
-    appLogoUrl: 'https://app.sky.money/images/sky.svg'
+    appLogoUrl: 'https://app.sky.money/images/sky.svg',
+    preference: { telemetry: false }
   }),
   coinbaseWallet({
     appName: 'sky.money',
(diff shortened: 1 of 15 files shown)


inclinedmind:
@coderabbitai full review

codewiththeflowxyz (approved):

PR #1584, opened 2026-05-23
Pendle: correctness fixes from code review

inclinedmind:
## Summary

Seven targeted fixes to the Pendle integration surfaced by a code review. All are bug fixes — no new features — and each ships with regression tests. Touches Pendle code only, except `6f983b08` which fixes a latent issue in the shared `useSequentialTransactionFlow` hook (no-op for non-Pendle callers).

## Fixes

| Commit | Severity | Fix |
|---|---|---|
| `2b9b5440` | High | Floor `apiMinOut` against `amountOut * (1 - slippage)` in `buildVerifiedArgs`. The verified-args pattern was passing through whatever `minOut` the `/convert` API returned, undermining slippage protection if the API ever returned an unexpected value. +10 tests. |
| `6f983b08` | High | Freeze sequential-tx args at `execute()` time instead of re-snapshotting while `currentIndex === 0`. Otherwise a quote refetch during the ~15s approve window could swap args between what the user reviewed and what the wallet asks them to sign. Affects Pendle + Curve stUSDS, no-op for other callers. |
| `7378eb81` | Medium | Drop NaN rows from `normalizePendlePnlRows` at the API boundary; tolerate `ptAmount` values with more fractional digits than the underlying's decimals via a safe stringifier. Previously a malformed PnL row would crash the activity modal inside `useMemo` with `parseUnits('NaN', n)`. +13 tests. |
| `83543a4e` | Medium | Value PT positions mark-to-market in the Balances widget using the impliedApy-derived discount. Previously displayed face-value-at-maturity, overstating positions by ~2–10% depending on time-to-maturity. Matches Pendle's UI and the rest of the Balances widget. +15 tests. |
| `aeb80900` | Medium | Keep the redeem-modal's stored `onConfirm` fresh across re-renders via a ref. Previously the closure captured `writeHook` at launch time, so a post-launch output-token change would still sign the original token. +1 regression test. |
| `50284f3d` | Low | Label matured-PT redemption as "Pendle redeem" instead of "Pendle pool" — post-expiry there's no AMM pool in the path. |
| `f9adee98` | — | Comment polish on the `executeRef` indirection. |

## Test plan

- [ ] `pnpm typecheck` from repo root
- [ ] `pnpm exec vitest run src/hooks/pendle src/modules/pendle src/widgets/PendleWidget` from `apps/webapp` — all green
- [ ] On Tenderly: open redeem modal on a matured PT, switch output between underlying / USDS / USDC — verify the displayed numbers and wallet popup args track the selection
- [ ] On Tenderly: open the Pendle buy/sell flow, advance to REVIEW, approve — verify the swap-tx args don't shift between review and the wallet popup

## Known but deferred

Two real but low-impact items intentionally left unfixed:

1. **`usePendleSlippage` doesn't sync state across sibling hook instances.** A user adjusting slippage on one matured-redeem card doesn't see it reflected in another card until that card's modal is reopened and the gear is touched. Self-healing per card; the displayed slippage in each modal still matches what's signed (no fund-safety implication). Mostly affects users with PT in multiple matured markets, which is uncommon.

2. **Slippage input is not debounced.** Typing in the custom slippage field fires a `/convert` per keystroke. Trade widget has the same pattern and hasn't been flagged as a problem in practice. Easy to add a debounce or commit-on-blur if telemetry shows excessive API hits.

## Architectural notes (non-blocking, worth knowing)

- **TransactionContext stores JSX + callbacks in state with an asymmetric `updateModalContent` API.** The redeem-modal `onConfirm` bug stemmed directly from `onConfirm` not being in the update set. Locally fixed via an `executeRef` ref-indirection; a render-prop refactor of `TransactionContext` would eliminate the smell at its source.
- **Pendle could adopt Trade's local-slippage-computation pattern.** Trade computes `sellAmountToSign` / `buyAmountToSign` client-side from the API's raw quote. Doing the same for Pendle (`minOut = amountOut * (1 - slippage)`) would obsolete the `apiMinOut` floor check, eliminate slippage-keystroke API hammering, and match Trade's architecture.

### DIFF
--- a/apps/webapp/src/hooks/pendle/buildVerifiedArgs.test.ts
+++ b/apps/webapp/src/hooks/pendle/buildVerifiedArgs.test.ts
@@ -91,7 +91,8 @@ const BUY_KNOWN = {
   outputToken: PT_USDG,
   underlyingToken: USDG,
   amountIn: 100_000_000n,
-  pinnedPendleSwap: PINNED_PENDLE_SWAP
+  pinnedPendleSwap: PINNED_PENDLE_SWAP,
+  slippage: 0.002 // matches the default 99.8M/100M apiMinOut/amountOut in makeBuyQuote
 };
 
 describe('buildVerifiedArgs — Buy', () => {
@@ -261,7 +262,8 @@ describe('buildVerifiedArgs — Withdraw', () => {
     outputToken: USDG,
     underlyingToken: USDG,
     amountIn: 100_000_000n,
-    pinnedPendleSwap: PINNED_PENDLE_SWAP
+    pinnedPendleSwap: PINNED_PENDLE_SWAP,
+    slippage: 0.005 // matches 99.5M/100M
   };
 
   function withdrawVerified(quote: PendleConvertQuote, known = WITHDRAW_KNOWN) {
@@ -317,6 +319,7 @@ describe('buildVerifiedArgs — Buy aggregator branch', () => {
     inputToken: USDS, // user-picked, NOT the underlying
     outputToken: PT_USDG,
     underlyingToken: USDG,
+    slippage: 0.002,
     amountIn: 100_000_000_000_000_000_000n, // 100 USDS (18 decimals)
     pinnedPendleSwap: PINNED_PENDLE_SWAP
   };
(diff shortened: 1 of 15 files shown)


inclinedmind:
@coderabbitai review

codewiththeflowxyz (approved):

iceanddust (approved):

PR #1593, opened 2026-05-27
[DO NOT MERGE] APP-247: drive react-hooks rules to error (incremental)

codewiththeflowxyz:
> **Status: DRAFT — DO NOT MERGE YET.**
> Base branch is intentionally `feature/app-175-pendle-integration` so this work tracks against the same code Pendle is shipping. Once the Pendle integration lands on `development`, the base will be **retargeted to `development`** and this PR will be marked ready.

Linear: [APP-247](https://linear.app/skybasestar/issue/APP-247/fix-outstanding-eslint-plugin-react-hooks-warningserrors)
Companion: [APP-227](https://linear.app/skybasestar/issue/APP-227/adopt-eslint-plugin-react-hooks-react-teams-official-plugin) (plugin adoption, already landed via #1539)
Unblocks: [APP-246](https://linear.app/skybasestar/issue/APP-246/install-and-enable-the-react-compiler) (React Compiler install)

## What this PR does

Drives the `eslint-plugin-react-hooks` v7 warnings introduced by #1539 down to zero, **one rule at a time**, in self-contained commits. After each rule's violations are cleared, the rule is promoted back to its preset default (`error`) by deleting its entry from the `TODO(APP-227)` block in `eslint.config.mjs` — so CI catches any regression from that point on.

Each commit follows the same shape: triage the rule's violations → fix the root causes (no per-line disables unless documented) → re-run lint to confirm 0 violations → promote the rule to `error` → verify `pnpm typecheck` + `pnpm -F webapp test` stay green.

## Rule-by-rule progress

| Rule | Initial count | Final count | Status | Commit |
|---|---:|---:|---|---|
| `rules-of-hooks` | 11 | **0** | ✅ promoted to `error` | `22f70a1e` |
| `purity` | 3 | **0** | ✅ promoted to `error` | `c069573e` |
| `immutability` | 8 | **0** | ✅ promoted to `error` | `4262a4be` |
| `preserve-manual-memoization` | 8 | **0** | ✅ promoted to `error` | `d83d0fee` |
| `static-components` | 9 | **0** | ✅ promoted to `error` | `3563ec5e` |
| `set-state-in-effect` | 83 | 83 | ⏳ pending — biggest bucket | — |
| `refs` | 35 | 35 | ⏳ pending | — |
| `exhaustive-deps` | 227 | 227 | not in scope (preset default is `warn`) | — |

**Cleared so far:** 39 violations across 5 rules.
**Remaining for promotion:** 2 rules (118 violations).
**TODO(APP-227) block** in `eslint.config.mjs` is down from 7 demoted rules to 2.

## Highlights / non-trivial fixes worth flagging for review

- **`useRewardsRate.ts` cache mutation (in immutability commit)** — was mutating the shared TanStack Query cache via `pricesData['LSMKR'] = lsMkrPriceData`. Every render of this hook injected `LSMKR` into the cache object that every other `usePrices()` consumer reads. Replaced with a local `pricesWithLsMkr = { ...pricesData, LSMKR: ... }`. **Real latent bug** — verified no other consumer reads `LSMKR` from prices, so the fix is safe; it just stops the cache pollution.
- **`Date.now()` in render → `useSyncExternalStore` (in purity commit)** — three hooks (`useBatchPendleConvert`, `useAllPendleUserAssets`, `useMerklRewards`) were calling `Date.now()` inside `useMemo` bodies. Replaced with React's sanctioned `useSyncExternalStore` pattern: `useBatchPendleConvert` arms a single `setTimeout` at the exact 90 s quote-TTL boundary; the other two poll on a 15 s interval. `STALE_QUOTE_ERROR` hoisted to module scope so `new Error()` doesn't run during render.
- **`TradeTransactionStatus.tsx` effect deps (in rules-of-hooks commit)** — moving two `useEffect`s above the early return required also adding `originToken` / `targetToken` to the second effect's deps array, otherwise the effect would no longer re-fire when tokens transition undefined → defined (a behavior regression that the lint rule alone wouldn't have caught).
- **`Repay.tsx` `calculateMaxRepayable` was a `useCallback` invoked immediately at render time (in preserve-manual-memoization commit)** — converted to `useMemo` of the return value. Same semantics, removes a layer of indirection, and cleared three knock-on memoization violations in the same file.

## Plan of action (remaining)

1. **`set-state-in-effect` (83 violations)** — the biggest bucket. These flag `setState(...)` calls inside `useEffect` that could trigger cascading renders. Common patterns: "mirror prop into state" (anti-pattern per React docs — compute during render instead), or "detect change to bump" (often refactorable to render-time `setState` with a guard, or eliminated entirely). Likely needs to be split across 2–3 sub-buckets (e.g., trade widget effects, balances effects, widget-context plumbing). Will use **test-first refactoring** for any effect that touches widget state or user-observable timing — write a characterization test before changing each behavior-sensitive site.
2. **`refs` (35 violations)** — `useRef` access patterns the rule disallows (e.g., reading `ref.current` during render, naming conventions). Mostly mechanical per #1539's notes (concentrated in 7 files: analytics banner, transaction context, sequential tx flow, etc.). No tests typically needed since refs are stable identity by definition.

## Companion work after this PR ships

Once `set-state-in-effect` and `refs` are cleared, the `TODO(APP-227)` block is empty and the v7 `recommended` preset's full ruleset is active at `error`. That unblocks [APP-246](https://linear.app/skybasestar/issue/APP-246) (React Compiler install) — the compiler will be able to optimize most components in the webapp because they now follow Rules of React strictly.

A small follow-up to consider: **retrofit characterization tests for the LSMKR cache fix and the Pendle quote-TTL boundary** — both were high-value bugs that would have benefited from a regression-locking test alongside the fix.

## Test plan

- [x] `pnpm install` clean (no new peer-dep warnings, no lockfile changes)
- [x] `pnpm lint` exits 0 (currently: 0 errors, 380 warnings — down from 406 before this PR)
- [x] `pnpm typecheck` clean after every commit
- [x] `pnpm -F webapp test` — 62/62 files, 265/265 tests pass
- [ ] `pnpm test:hooks` (vnet-backed) — not yet run; targets are pure-JS / pure-component logic with no on-chain interaction, but worth running before retargeting to `development`
- [ ] CI green
- [ ] Manual smoke test in dev environment before retarget to `development`

## Why DRAFT / DO NOT MERGE

Cleanup work isn't blocking Pendle's Monday (2026-06-01) ship. Two reasons to hold:
1. Keeps the Pendle branch's review surface focused on Pendle-specific changes.
2. Once Pendle merges to `development`, this PR rebases cleanly onto `development` and ships separately. Avoids merge-conflict churn in either direction.

If Pendle slips significantly past Monday, this PR will be split per the original plan (cleanup that touches pre-existing `development` code → small PRs straight to `development`; cleanup of Pendle-specific net-new code → stays on this branch).

### DIFF
--- a/apps/webapp/src/hooks/morpho/useMerklRewards.ts
+++ b/apps/webapp/src/hooks/morpho/useMerklRewards.ts
@@ -1,6 +1,6 @@
 import { useQuery, useQueryClient } from '@tanstack/react-query';
 import { useChainId, useConnection, useReadContracts } from 'wagmi';
-import { useCallback, useMemo } from 'react';
+import { useCallback, useMemo, useSyncExternalStore } from 'react';
 import { TRUST_LEVELS, TrustLevelEnum } from '../constants';
 import { ReadHook } from '../hooks';
 import { isTestnetId, formatBigInt } from '@/utils';
@@ -309,12 +309,19 @@ export function useMerklRewards(): MerklRewardsHook {
     }
   });
 
+  // Subscribe to wall-clock time via React's sanctioned external-store API.
+  // 15s precision: filter clears within 15s of the 5-minute recent-claim mark.
+  const subscribeTick = useCallback((notify: () => void) => {
+    const id = setInterval(notify, 15_000);
+    return () => clearInterval(id);
+  }, []);
+  const now = useSyncExternalStore(subscribeTick, () => Date.now(), () => 0);
+
   // Filter out tokens that were recently claimed on-chain but still show in the API
   const data = useMemo(() => {
     if (!apiData) return undefined;
     if (!claimedData || claimedData.length === 0) return apiData;
 
-    const now = Date.now();
     const filteredRewards = apiData.rewards.filter((_reward, index) => {
       const result = claimedData[index];
       if (!result || result.status === 'failure') return true;
@@ -330,7 +337,8 @@ export function useMerklRewards(): MerklRewardsHook {
       rewards: filteredRewards,
       hasClaimableRewards: filteredRewards.length > 0
     };
-  }, [apiData, claimedData]);
+  }, [apiData, claimedData, now]);
+
 
   const mutate = useCallback(() => {
     if (!userAddress) return;
--- a/apps/webapp/src/hooks/pendle/useAllPendleUserAssets.ts
+++ b/apps/webapp/src/hooks/pendle/useAllPendleUserAssets.ts
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useCallback, useMemo, useSyncExternalStore } from 'react';
 import { TRUST_LEVELS, TrustLevelEnum } from '../constants';
 import { usePrices } from '../prices/usePrices';
 import { usePendleUserPtBalances } from './usePendleUserPtBalances';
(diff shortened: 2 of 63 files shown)


iceanddust:
Closing in favour of #1955 (lint gate config: prettier in CI, warning ratchet, three react-hooks rules back to `error`) and #1956 (refs, exhaustive-deps and no-console fixed and promoted to `error`). This branch is four months behind development: roughly half its files were deleted with the redesign (`src/widgets`, WidgetNavigation, seal, LinkedActionCard) and the rest overlaps the two PRs above.

The `set-state-in-effect` work here (useIsTouchDevice, useClipboard, ConnectModal, ConnectedContext, TermsModalContext) will be reviewed for reuse when that rule gets its own PR.

PR #1609, opened 2026-06-01
Add the Tether Savings (sUSDT) vault

codewiththeflowxyz:
### What does this PR do?

Adds the **Tether Savings (sUSDT)** vault to the Vaults tab as a new provider alongside the existing Morpho vaults. A small provider abstraction lets both share the same vault machinery, with the existing Morpho vaults unchanged in behavior.

**Working now (on-chain):**
- The vault appears in the Vaults list and "My vaults"; users can deposit USDT to receive sUSDT and withdraw/redeem, end to end.
- TVL, exchange-rate-adjusted balances, and the deposit cap + withdrawal limit are read from the contract, so the input is clamped to amounts the contract will accept.
- "Powered by Spark" branding distinguishes it from the Morpho vaults.

**Data-dependent (rate/APY, allocations, history chart):**
- Wired through a Spark data source that reads from a placeholder URL until the live endpoint is delivered. Until then these show a clean empty state and TVL falls back to the on-chain read.

### Testing steps:

1. Open the **Vaults** tab → **Tether Savings**.
2. Confirm the TVL, and (when connected with USDT) your balance display.
3. Deposit USDT → receive sUSDT; verify the input is capped to your balance / the vault's remaining capacity.
4. Withdraw / max-withdraw (redeem) sUSDT back to USDT.
5. Confirm the "Powered by Spark" badge, and that the existing Morpho vaults are unchanged.

Automated coverage: unit tests for the limits math, the Spark data normalization, and provider dispatch; an end-to-end deposit/withdraw test; component tests for the allocations and chart rendering.

### TODOs / follow-ups

- [ ] Point `SPARK_VAULT_API_URL` at the live Spark endpoint once delivered (currently a placeholder) — this enables the rate/APY, allocations, and history chart.
- [ ] Confirm rate semantics (net vs gross) against the real payload — localized to `normalizeSparkVaultPayload`.
- [ ] Refresh the e2e / hooks test fork past the vault's deployment block so on-chain coverage runs against the real sUSDT vault (the address has no bytecode at the current pinned block, so the on-chain limit test runs against an equivalent Morpho ERC-4626 vault).
- [ ] Wire a per-user transaction-history source for the vault when one exists (the history section is currently limited to the other vaults).
- [ ] Consider labelling vault events per provider (they currently disambiguate via product / address).
- [ ] Optional: finish renaming the shared widget/config off the legacy `Morpho*` names (kept as thin aliases to keep this diff small).
- [ ] Minor UX: the supply balance line reads "0" both when the wallet is empty and when the deposit cap is reached — distinguish the two.


apps/webapp/src/hooks/vaults/spark/constants.ts
```diff
@@ -0,0 +1,52 @@
...
+import { TOKENS } from '../../tokens/tokens.constants';
+import { VaultConfig } from '../types';
+
+/**
+ * Tether Savings (sUSDT) `SparkVault` — the only Spark contract the webapp touches.
+ * Verified ERC-4626 proxy; upgrade authority = Sky governance Pause Proxy (APP-266,
+ * address pre-cleared). Underlying asset is USDT (6 decimals). Same address on
+ * mainnet and the Tenderly fork.
+ */
+export const SPARK_USDT_VAULT_ADDRESS = '0x74cb54e082411cfCAEADb00a0765625B10410DAa' as const;
```
bitsforpeace: Why do we define this here rather than in contracts file?

bitsforpeace (approved):

PR #1647, opened 2026-06-10
fix(vaults): sUSDT vault review follow-ups

bitsforpeace:
Stacked on top of #1609 — small, low-risk fixes surfaced while reviewing it. Targets the `feature/app-266-tether-savings-susdt-vault` branch so it can land into the feature PR before that merges to `development`.

### Changes

- **Sky rate tooltip on the balances suggested-actions** — `RATE_TOOLTIP_TYPES.sparkVault` mapped to the `'morpho'` tooltip, so the Tether Savings suggested-action rendered the "Sky as vault curator on Morpho… third-party lending protocol" copy. Switched to the existing dedicated `'sky'` type (`susdt-vault-rate`), matching `SparkVaultRate` and `VaultsBalanceCard`. *(user-visible)*
- **Sky balance-card data source** — `VaultBalanceDetails` called `useMorphoVaultOnChainData({ vaultAddress })` without forwarding `provider`, so the Sky vault's data-source transparency label read "Morpho Vault Contract". Now forwards `provider`. *(cosmetic; balances/links were already correct)*
- **e2e spec honesty** — `vaults-spark.spec.ts` is now `test.describe.skip` with a documented reason: the Tenderly fork is pinned before the vault's deploy block (no bytecode → reverts) and funds no USDT, and the spec isn't in `playwright-parallel.config.ts` `testMatch` so it never ran. Comment explains exactly how to re-enable.
- **Dead code** — removed `VaultPoweredBy.tsx` (never imported; the live powered-by UI is `VaultPoweredByBadge`; its string was also un-i18n'd).

### Notes
- No behavior change for Morpho vaults.
- Changes are type-safe by inspection (`'sky'` is a valid `PopoverTooltipType`; `provider` is in scope and `useErc4626VaultData` accepts it). I could not run the full typecheck/lint in the isolated review worktree (no `node_modules`) — worth letting CI confirm.
- Remaining review items left for the feature PR / follow-ups: max-redeem can revert under a liquidity-constrained vault (pre-existing, safe failure), finishing the `spark`→`sky` rename, and the bundled APP-279 corpus-sync/copy changes in #1609's diff.


apps/webapp/src/test/e2e/tests/vaults-spark.spec.ts
```diff
@@ -7,10 +7,19 @@ import { connectMockWalletAndAcceptTerms } from '../utils/connectMockWalletAndAc
 // NOTE: assumes the Tenderly fork funds the test account with USDT (operator
 // runs e2e). Provider-aware selectors: `sky-vault-stats-card`,
 // `supply-input-sky`, `withdraw-input-sky`.
+//
+// SKIPPED: this spec is not yet runnable and is intentionally not advertised as
```
codewiththeflowxyz: Why is this? I think the tests passed in other PRs related to this feature

codewiththeflowxyz (approved):

PR #1664, opened 2026-06-12
feat(earn): C2 — Earn Opportunities section

iceanddust:
## Summary

Implements ticket **C2 scoped to the "Earn Opportunities" section** (hero + featured cards deferred): the real marketplace table replaces the C1 debug table on `/earn`, fed by `useEarnMarketplace`. Stacked on `redesign/c1-earn-registry` (#1663) — retarget when that merges.

### What's here

**First L1 composites in `src/components/product/`** (per the conventions doc layer model — layout/props only, icons arrive as ReactNode slots, no data/router/engine imports):
- `EarnTable` — sortable column headers (active column highlighted, `aria-sort`, per-column default direction: numeric desc, rest asc), card-style rows on the `container` surface with a `containerDark` hover, per-cell skeletons while a row's sources load, keyboard-accessible row selection.
- `EarnTableFilters` — risk pills (existing Button `chip`/`pill` variants) + the three dropdowns (networks / stablecoins with token icons / products) on the V2 popover treatment: `containerDark` surface, hover rows, selected row shown by a prominent background.
- `RiskMeter` — three-segment tier indicator (low green / moderate orange / advanced red).

**L2 wiring in `modules/earn/`**: pure `earnTableState` helpers (sanitize/filter/sort, unit-tested), `useEarnTableState`, and the rewritten `EarnPage`. Rows navigate to their `detailPath` retaining `network`.

**Containerless full-width destinations**: `AppContainer` gains a `card | bare` variant — `bare` keeps the identical sizing/scroll box (header keeps its place while the table scrolls) minus the card chrome. Applies to `/earn` and `/portfolio`.

**Registry addition (C1 layer)**: rows now declare `supplyTokens` — Savings supplies USDS, Pendle accepts USDS/USDC/sUSDS (hardcoded with a comment; flag if a canonical Pendle input-token constant exists), vaults their asset, stUSDS USDS. The "Supply:" icons and the stablecoin filter consume them.

**One primitive extension** (extend `components/ui`, don't fork): `SelectItem` gains opt-in `hideIndicator` — V2 dropdowns show selection by row background, not checkmark. The only other Select consumer (`Chart.tsx`) keeps the default.

### Product decisions encoded
- **Filters persist to localStorage** (`earnOpportunitiesFilters`), sanitized against the offered options on load; **sort is ephemeral**, defaulting to **risk ascending** each visit (provisional). ⚠️ This is a deliberate deviation from the C2 AC's "filter state URL-controlled and deep-linkable" — decided 2026-06-12 (the URL-param governance cost wasn't worth it now; URL params can later seed the same state without rework). Please record in the decision log.
- **Risk pills multi-select; none selected = all tiers shown.**

### Testing
- Unit: `earnTableState.test.ts` (sanitize drops stale values, filter semantics, stable sort with undefined-last, per-column default directions); registry tests extended for `supplyTokens`; `AppShell.test` asserts the containerless full-width branch. Fast suite 498/498; typecheck/lint clean.
- Manual (Playwright, dark + light): filters/sort/row-nav exercised — Advanced pill isolates stUSDS, TVL first-click sorts desc, filters survive reload with pills pressed, row click lands on `/earn/savings` keeping `network`; hover states and dropdown styling captured.
- **V2 e2e for filters/sort still owed** — local `pnpm vnet:fork` returns Forbidden (Tenderly key); follow-up once CI/key works.

### Drive-by findings (not fixed here)
- `bg-bgHover` is an **undefined token** — TopNav's MoreMenu items have a silent no-op hover (same one-liner when B5 re-skins chrome).
- `--color-secondaryHover` is defined identical to `--color-secondary`, making Button chip hovers invisible — looks like A2 token-audit territory.


apps/webapp/src/modules/app/shell/TopNav.test.tsx
```diff
@@ -126,7 +126,7 @@ function renderTopNav(initialPath = '/') {
 
   render(
     <I18nWidgetProvider locale="en">
-      {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
+      {}
```
bitsforpeace: Is the empty bracket necessary?

bitsforpeace (approved):
Looking good. Just left a small question.

apps/webapp/src/modules/app/shell/TopNav.test.tsx
iceanddust: Good catch, I just removed it

codewiththeflowxyz (approved):

PR #1679, opened 2026-06-17
Gate balances suggested-action cards off of geo-config module flags

inclinedmind:
## What

Replaces the hardcoded `isRegionRestricted` whitelist in `BalancesDetails` with the per-module `isModuleEnabled` filtering that already exists in `BalancesSuggestedActions`.

Previously the "Earn with your Stables" section passed a hardcoded `restrictedModules={['morpho','fixedYield']}` whitelist (gated on the blunt `isRegionRestricted` boolean), while `BalancesSuggestedActions` only mapped `trade` to its geo-config module. This completes the `action module -> ModuleId` mapping so each suggested-action card is gated by its own fine-grained geo-config module flag (`vaults`, `rewards`, `savings`, `fixed`, `expert`, `stake`, `upgrade`).

## Why

- Removes a hardcoded, region-boolean-driven restriction in favor of the server-controlled per-module config.
- The `restrictedModules` prop name was misleading — it was actually a *whitelist* (kept the listed modules), the inverse of what the name implied.
- Makes card visibility consistent with how widget navigation is already gated.

## Behavior

**No behavior change under the current config.** The old whitelist kept `morpho` + `fixedYield` and dropped rewards/savings/expert — exactly what the module flags produce when restricted (`vaults`/`fixed` enabled; `rewards`/`savings`/`expert` disabled).

Note: completing the mapping also brings the SKY card (`stake`) and the upgrade token cards (`upgrade`) under module control. Harmless today (both enabled), but the geo-config can now hide them too — the consistent behavior.

## Testing

- `pnpm -F webapp typecheck` passes
- eslint clean on both files
- Existing `BalancesSuggestedActions.test.tsx` (4 tests) passes


inclinedmind:
@coderabbitai review

iceanddust (approved):

bitsforpeace (approved):

PR #1681, opened 2026-06-19
D3 · Reusable savings supply flow: trigger + shared form model (Portfolio in-place supply)

codewiththeflowxyz:
### What does this PR do?

Makes the Savings supply/withdraw flow reusable by any surface, and uses it from the Portfolio **Supplied** carousel so a position card's **Supply** button opens the modal in place instead of navigating to the product page. Calldata and the savings engine hooks are unchanged.

**Reusable trigger**
- New `useSavingsModal()` returns `{ openSupply, openWithdraw }`, each assembling the `launch()` config (title, subtitles, `entry`, `backgroundContent`, session id) so callers no longer copy ~20 lines of config per entry point.
- Each opener takes an optional `{ amount?, token? }` preset, seeded into the form's initial state on launch (the host already remounts per launch, so no effect is needed).
- `SavingsPositionCard` consumes the hook instead of inlining the two configs; Portfolio reuses the same `openSupply`.

**Shared form model**
- New headless `useSavingsTransactionForm()` owns the supply/withdraw form: amount/Max/token state, origin-token resolution per flow + network, wallet/sUSDS balance reads, L2 PSM preview reads, the spend gate, and the `useSavingsLaunch` engine-param mapping.
- The editable modal body (`SavingsModalForm`) and the inline no-position panel (`SavingsSupplyWithdrawPanel`) each previously carried their own copy of all of that; both are now thin presentation over the one hook, keeping only their own chrome (the panel's tabs + review-modal CTA; the modal's entry-slot portal + `updateModalContent` sync + before/after rows).
- Net: the supply form logic is single-sourced across all three surfaces — the no-position inline panel, the has-position editable modal, and the Portfolio in-place modal. ~350 lines of duplication removed; behavior and calldata unchanged.

**Portfolio carousel**
- `PositionCard`'s single `onSelect` is split into `onSupply` / `onManage`, so the caller owns each button.
- A `usePortfolioSupplyActions` resolver maps a position's `kind` to its in-place supply opener (or `undefined` → navigate). Adding a future product is a single case in the resolver, never a branch in the carousel JSX.
- The savings card opens in place **only when its position is on the connected chain**. An off-chain card — e.g. a Base-badged savings position while the wallet is on mainnet — falls back to navigation, so the modal never supplies on a chain other than the one the card represents.

No analytics/telemetry changes.

### Testing steps:

Automated (Vitest): `useSavingsModal` config + preset; `usePortfolioSupplyActions` kind + connected-chain dispatch; `PortfolioPositionsSection` routing. The form-model extraction is behavior-preserving, so the existing `SavingsModalForm` and `SavingsSupplyWithdrawPanel` suites pass unchanged (the modal suite gains one mock for the now-shared, gated-off `useReadSavingsUsds` read). Full savings suite 128/128; typecheck and lint clean.

Manual (`pnpm -F webapp dev:mock` → Portfolio → **Supplied** tab):
- Sky Savings Rate card **Supply** while on the connected chain → opens the modal in place (no navigation); a supply succeeds.
- Sky Savings Rate card **Manage** → navigates to the savings product page.
- Any non-savings card **Supply** / **Manage** → navigate (unchanged).
- Set the network filter to a chain the wallet is **not** connected to → that savings card's **Supply** navigates to the product page instead of opening a mismatched modal.

Also exercise the savings product page itself (same shared form model now backs all of these):
- No-position **Supply** card → still inline amount entry → read-only review modal.
- Has-position **Supply** / **Withdraw** → still open the editable modal.

### Notes / known follow-ups

- **Post-supply refresh:** opening from Portfolio passes no `onSuccess` (the marketplace query exposes no refetch handle), so the card refreshes on the next window-focus refetch rather than instantly. Deferred.
- `usePortfolioSupplyActions` runs above the section's early returns, so the idle/empty/loading paths now transitively require a `TransactionProvider` (always present today via `App.tsx`). Latent only.
- The inline panel's uncontrolled Supply/Withdraw tab mode now has no production consumer (`SavingsSupplyCard` always passes `flow="supply"`); it's preserved for its test but is a candidate to prune.

### DIFF
--- a/apps/webapp/src/modules/portfolio/components/PortfolioPositionsSection.test.tsx
+++ b/apps/webapp/src/modules/portfolio/components/PortfolioPositionsSection.test.tsx
@@ -1,0 +1,145 @@
+import { i18n } from '@lingui/core';
+import { I18nProvider } from '@lingui/react';
+import { render, screen, within, cleanup, fireEvent } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { PortfolioPositionsSection } from './PortfolioPositionsSection';
+import type { SuppliedPosition, SuppliedView } from '../helpers/suppliedView';
+import type { IdleView } from '../helpers/idleView';
+
+i18n.load('en', {});
+i18n.activate('en');
+
+const h = vi.hoisted(() => ({ navigate: vi.fn(), openSupply: vi.fn(), openWithdraw: vi.fn(), chainId: 1 }));
+
+vi.mock('@tanstack/react-router', () => ({ useNavigate: () => h.navigate }));
+
+// The resolver reads the connected chain to gate in-place supply; keep real wagmi
+// exports, override only useChainId.
+vi.mock('wagmi', async importOriginal => {
+  const actual = await importOriginal<typeof import('wagmi')>();
+  return { ...actual, useChainId: () => h.chainId };
+});
+
+// The savings trigger under test — capture its opener.
+vi.mock('@/modules/savings/hooks/useSavingsModal', () => ({
+  useSavingsModal: () => ({ openSupply: h.openSupply, openWithdraw: h.openWithdraw })
+}));
+
+// Carousel → passthroughs so the cards render flat (no embla/DOM coupling).
+vi.mock('@/components/ui/carousel', () => ({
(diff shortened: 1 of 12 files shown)


iceanddust:
@coderabbitai review

iceanddust (approved):

PR #10, opened 2025-01-21
arbitrum feature branch

inclinedmind:

PR #1690, opened 2026-06-24
feat(terms): version-aware terms re-prompt — surface latest version (APP-330)

bitsforpeace:
## What

Frontend half of version-aware wallet terms (APP-330, parent APP-152). Threads the `latestVersion` returned by `/terms-acceptance/check` through `checkTermsWithRetry` → `ConnectedContext` → `TermsModal`, populating `TermsDialog`'s **existing** `termsVersion` prop (which already renders "Terms version: {version}").

The terms gate already keys purely off the `check` result (`ConnectedContext` sets `hasAcceptedTerms` from it on every connect — no persistence), so no gating change is needed: once the backend is version-aware, a stale wallet's `check` returns `false` and the existing modal flow re-prompts it automatically. This PR surfaces *which* version is being requested and locks the behavior with tests.

## ⚠️ Depends on two other PRs

This is the last of three and is **inert until the other two ship** — on its own, the backend still returns `termsAccepted: true` for any prior acceptance, so nothing re-prompts.

1. **DB** — `sky-money-supabase` #10 (APP-328): adds the `version` column + composite `(address, version)` PK to `terms-acceptance`.
2. **Backend** — `api-workers` #103 (APP-329): makes `/terms-acceptance/check` and `/add` version-aware and returns `latestVersion`.
3. **Frontend** — this PR (APP-330).

Roll out in that order: DB → api-workers → this. End-to-end staging verification requires #10 merged to the supabase `staging` branch and #103 deployed to `staging-api.sky.money`.

## Expected behavior once all three are implemented

| Scenario | `check` result | UX |
| --- | --- | --- |
| New wallet, never accepted | `termsAccepted: false` | Terms modal opens (shows current version). Sign → `/add` stores acceptance at current version → `check` true → modal closes. |
| Wallet already accepted the **current** version | `termsAccepted: true` | No prompt. (Existing acceptances are backfilled to the current version by APP-328, so the rollout itself does **not** re-prompt anyone.) |
| Terms version bumped, wallet accepted an **older** version | `termsAccepted: false` | On next connect/load the modal **re-opens** showing the new version. Re-sign → `/add` inserts a **new** row keyed on `(address, newVersion)`; the prior signature is retained as an audit trail → `check` true → modal closes and stays closed. |
| VPN / restricted region / sanctioned | 403 `accessDenied` | Blocked UI (unchanged). |

Re-prompting happens on the next connect/load (the check isn't re-run mid-session), which matches the intended behavior.

## Tests

- `ConnectedContext.test.tsx` (new): `termsAccepted:false` → wallet stays gated (re-prompt); `termsAccepted:true` → gate opens. Also asserts `latestVersion` is surfaced.
- `checkTermsWithRetry.test.ts`: locks the `latestVersion` passthrough.
- 16 tests pass; typecheck, lint, prettier clean.

## Note

Per the agreed scope this is frontend-only — no "your terms were updated" differentiation messaging (that would need an additive backend signal to tell stale from never-accepted). The returning user sees the standard modal with the version label. Easy follow-up if we want richer copy.

Closes APP-330.

### DIFF
--- a/apps/webapp/src/modules/ui/components/TermsModal.tsx
+++ b/apps/webapp/src/modules/ui/components/TermsModal.tsx
@@ -23,7 +23,8 @@ export function TermsModal() {
     termsCheckError,
     retryTermsCheck,
     isConnectedAndAcceptedTerms,
-    setHasAcceptedTerms
+    setHasAcceptedTerms,
+    latestTermsVersion
   } = useConnectedContext();
   const [isChecked, setIsChecked] = useState(false);
   const [signStatus, setSignStatus] = useState<'idle' | 'loading' | 'signing' | 'error'>('idle');
@@ -221,6 +222,7 @@ export function TermsModal() {
       isOpen={isModalOpen}
       onOpenChange={handleOpenChange}
       title={<Trans>Legal Terms</Trans>}
+      termsVersion={latestTermsVersion}
       content={termsContent}
       additionalContent={checkboxContent}
       customError={errorContent}
--- a/apps/webapp/src/modules/ui/context/ConnectedContext.test.tsx
+++ b/apps/webapp/src/modules/ui/context/ConnectedContext.test.tsx
@@ -1,0 +1,104 @@
+import { render, screen, waitFor } from '@testing-library/react';
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+// Shared spies/state between the mocks and assertions.
+const mocks = vi.hoisted(() => ({
+  checkTermsWithRetry: vi.fn(),
+  trackVpnCheckCompleted: vi.fn(),
+  reportError: vi.fn(),
+  isPrivateDeployment: vi.fn(() => false)
+}));
+
+vi.mock('wagmi', async importOriginal => {
+  const actual = await importOriginal<typeof import('wagmi')>();
+  return {
+    ...actual,
+    useChainId: () => 1,
+    useConnection: () => ({ isConnected: true, address: '0x1234567890123456789012345678901234567890' })
+  };
+});
+
+vi.mock('@/hooks', () => ({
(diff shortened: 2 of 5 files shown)


iceanddust:
Closing in favour of #1832 (APP-499), which rewrites both files this PR touched.

**Why closed rather than rebased.** This branch is based on `development` and was last touched 24 June. `ConnectedContext.tsx` has moved 19 lines on `app-redesign` since, and the whole T&C flow now targets `app-redesign` via an integration branch (decided 13 Aug, recorded in APP-503). More to the point, the ~15 lines of production code here were written against the old contract — `res.termsAccepted`, a single boolean — which api-workers #112 replaced with `{ accepted, signedForCurrentVersion, latestVersion }`, plus `messageToSign` in #113. That code would have been rewritten in the rebase regardless.

**Where each piece went:**

- **`ConnectedContext.test.tsx` (+104 lines)** — the bulk of this PR. Cherry-picked verbatim into #1832 as its own commit (`12fc6a1`) so the provenance survives, then adapted to the two-phase contract in the following commit. It now covers each half of the AND gate, the address switch, the version bump, and the write ordering, on top of the version-aware cases it started with.
- **`latestVersion` passthrough** — carried over. It is required now rather than optional: it keys the localStorage flag, so a response without it is an error instead of a silently version-less gate.
- **`termsVersion` threaded into the dialog** — carried over. `TermsDialog` already renders it; `TermsModal` now passes `latestTermsVersion`. The footer copy it feeds ("Terms of Use version [X.X], effective [date]") belongs to C4/APP-500, so only the plumbing lands in #1832.
- **The `/add` payload and `VITE_TERMS_MESSAGE_TO_SIGN`** — deliberately not carried. Phase A now posts `{ address }` with no signature, and the signing message moved into the worker (APP-508), so the env var is deleted rather than threaded.

Thanks — the test file saved real work.

PR #1701, opened 2026-07-02
Fixed-yield market detail page on the product template

codewiththeflowxyz:
## Summary

- New `/earn/fixed/:slug` route renders Pendle markets as a product detail page: historical rate/TVL chart, position card, maturity progress, details grid, About section, and transactions table. The legacy market-address route redirects to the slug route.
- Supply/withdraw run through the shared transaction modal with a new `SlippageMenu` primitive and per-flow persisted slippage. Step indicators and the bundle-transactions toggle now work like the savings and vault modals, and batching follows the user's toggle plus wallet support.
- About section matches the design: intro with a worked example computed from the live implied APY and remaining term, plus three icon accordion rows.
- The powered-by badge uses the brand Pendle mark at the design's size.
- Maturity gating and the convert-call verification allowlist are unchanged.

## Testing

- Unit suites for the pendle module (73 tests), plus ui/savings/morpho suites (256 total) pass; typecheck and lint clean.
- Manually verified on a Tenderly fork: supply and withdraw end to end (sequential and batched wallets), live position/balance updates without reload, per-flow slippage, chart toggles and timeframes, and the legacy route redirect.


apps/webapp/src/hooks/earn/earnProducts.ts
```diff
@@ -101,7 +101,7 @@ export function buildEarnProducts(
     supplyTokens: [TOKENS.usds.symbol, TOKENS.usdc.symbol, TOKENS.susds.symbol],
     risk: DEFAULT_RISK_TIER,
     networks: productNetworks(Intent.FIXED_INTENT, familyChainIds),
-    detailPath: intentToPath(Intent.FIXED_INTENT, `market/${market.marketAddress}`),
+    detailPath: intentToPath(Intent.FIXED_INTENT, market.slug),
```
iceanddust, line 104: I'm wondering about this change... At first it makes sense to me, but I'm thinking on a possible edge case in which we could have multiple PT markets of a single token (e.g. multiple PT-sUSDS fixed yield markets because one of them is close to expiration). Maybe we can keep market.slug but still attach the contract address?

iceanddust (approved):
LGTM, the only minor thing is that it looks like we still need to wire the Pendle supply modal launcher to the portfolio page's Pendle supply button.

apps/webapp/src/hooks/earn/earnProducts.ts
codewiththeflowxyz: Slugs are hand-authored per market entry and uniqueness across `PENDLE_MARKETS` is test-enforced, so a second live PT-sUSDS market would ship with its own slug (e.g. `pt-susds-may2027`, the config type doc calls out disambiguating by maturity). The address also still works as an identifier via the `/earn/fixed/market/:address` redirect; it just doesn't appear in the canonical URL. Keeping the route slug-only avoids reintroducing addresses into URLs and the slug/address mismatch case.

PR #1711, opened 2026-07-07
APP-309: E2 Convert swap surface — PSM-only page-as-widget

iceanddust:
## Summary

Rebuilds `/convert` as the E2 **page-as-widget** swap surface per the Figma UI file (frames `486:31193` / `486:31219` / `486:32223`), driving PSM 1:1 conversions through the shared `launch()` transaction modal. Linear: [APP-309](https://linear.app/ekliptyka/issue/APP-309).

**Scope (narrowed from the original ticket, agreed with product):**
- **PSM USDC↔USDS only**, on all 5 chains — mainnet via `UsdsPsmWrapper`, Base/Arbitrum/Optimism/Unichain via PSM3. No sUSDS, no CoW tokens.
- **Trade and Upgrade become unreachable** (routes redirect to `/convert`, More-menu Upgrade shortcut removed). Their modules/widgets stay parked in the tree pending the restore-vs-retire decision; Trade returns in E3 on the async-order contract.
- The ticket's blocking decision (module selector vs auto-routing) is mooted by this scope: two tokens, one engine, a flip button.

## How it works

- **Engine untouched**: `usePsmConversion` (+ helpers/tests) relocated verbatim from `widgets/PsmConversionWidget/hooks/` to `modules/convert/hooks/`. It owns all calldata, allowance, batching, liquidity/halted/fee guards.
- **Review-first launch** (pendle-redeem precedent): the swap form lives on the page and stays mounted under the overlay, so `launch()` carries only the read-only review body + steps + onConfirm — no `entry`/`backgroundContent`. One engine instance feeds both preview and `onConfirm`, so the reviewed figures and calldata cannot diverge.
- **Analytics parity** via `TransactionConfig.analytics` (`widgetName: 'convert'`) — the transaction context fires review-viewed/lifecycle events, keeping the legacy widget's event identity.
- **Deep links**: `?source_token=USDC|USDS` preselects direction; `/convert/psm?source_token=…` redirects preserving it; legacy `?widget=trade|upgrade|convert` URLs land directly on `/convert`.

## Notable deletions
- `PsmConversionWidget` (UI) + `ConvertPanes`/`ConvertWidgetPane`/`PsmConversionWidgetPane`/details/FAQ — superseded by the new page (same-PR deletion per D6/D7 precedent). Its flow/action/screen enums moved into the shared `widgetState` contract.
- `ConvertIntentMapping`, `CONVERT_MODULE_SELECTED` analytics, convert nav sub-items, the orchestrator's L2-upgrade redirect, and the `convertIntent` axis of `validateSearchParams`.

## Known stubs / follow-ups
- **Network fee row shows "–"** — no gas-estimate hook exists anywhere yet (same stub as savings/pendle modals).
- **No Details/FAQ section** — the Figma shows none; flagging since the legacy surface had both.
- Success toast is title-only ("USDC converted!"); the toast contract has no second line for the amount summary shown in Figma.
- E2E: **not rewritten** (branch convention — `pendle.spec.ts` is the only rewritten-V2 spec and is itself fully skipped). `e2e-migration.md` rows annotated: PSM specs need a full V2 rewrite against `convert-*` testids; trade/upgrade specs parked.

## Gates
- typecheck ✓ · lint 0 errors ✓ · unit 819/819 ✓ · vnet hooks suite 491/491 ✓ · build ✓
- Fork browser QA in progress (results will be posted here).


iceanddust:
@coderabbitai review

apps/webapp/src/modules/convert/hooks/useConvertLaunch.tsx
```diff
@@ -0,0 +1,196 @@
...
+  ]);
+
+  // Keep the open modal's gating + preview live while it still sits on the
+  // review screen (amounts can't change mid-review — the form is under the
+  // overlay — but prepared/allowance state can). Once Confirm fires the content
+  // freezes: the post-success form reset must not blank the executed amounts
+  // on the wallet/status screens.
+  useEffect(() => {
+    if (!isModalOpen || txStatus !== TxStatus.IDLE) return;
+    updateModalContent(sessionId, { transactionContent, transactionScreenContent, confirmDisabled, steps });
```
codewiththeflowxyz, line 202: **High — infinite render loop / "Maximum update depth exceeded" while the Review modal is open**

`transactionContent` (L102) and `transactionScreenContent` (L120) are rebuilt as fresh JSX on **every render**, then used as dependencies of this live-update effect. Once the review modal is open on the IDLE screen (and the page stays mounted under the overlay, by design), this loops:

1. Effect fires → `updateModalContent` → `setActiveConfig(prev => ({ ...prev, ...rest }))` in `TransactionContext.tsx` — **always a new object**.
2. Provider re-renders. Its context `value` is built inline and **not memoized**, so every consumer re-renders.
3. `ConvertPage` consumes the context (via `useConvertLaunch → useTransaction`) → re-renders → **new** `transactionContent` / `transactionScreenContent` references.
4. Effect deps changed → effect fires again → back to step 1.

This is the canonical "setState in an effect whose deps change every render" pattern, so React bails with **"Maximum update depth exceeded"** and the Review modal fails to render.

This is a regression from the cited precedent: `usePendleRedeemModal.tsx` wraps `transactionContent` in `useMemo` precisely so the effect deps are stable — this hook dropped that memoization.

**Why the tests don't catch it:** `useConvertLaunch.test.tsx` mocks `useTransaction`, so `updateModalContent` is a no-op `vi.fn()` — the setState→re-render feedback cycle that drives the loop doesn't exist under test. It only manifests with the real provider in the browser.

**Fix:** wrap both `transactionContent` and `transactionScreenContent` in `useMemo`, keyed on their actual inputs (`originSymbol`, `targetSymbol`, `amount`, `conversion.targetAmount`, decimals, `chainId`, `networkName`), matching the Pendle precedent.

codewiththeflowxyz (approved):

PR #1731, opened 2026-07-15
Nav Menu dropdown per the DS comp: switch rows for bundling & dark mode, iconed legal links

codewiththeflowxyz:
### What does this PR do?

image

Rebuilds the top-nav hamburger (More) menu to the design-system Navbar Menu comp (DS file, dropdown node 5069:27495):

- **Panel**: 274px glass panel — `bg-secondary` over a 100px backdrop blur, 24px radius, 20px padding, 24px between sections, hairline `border-secondary` divider.
- **Bundle transactions** (`BatchTransactionsToggle` rewritten): 16px zap glyph + Label 5 + S-size DS `Switch`, with the 216px Body-7 explainer underneath. The Legal Notice link and the unsupported-wallet notice (+ supporting-wallets link) carry over from the old tooltip — the comp omits them, but they're disclosures, so they stay; when a connected wallet lacks batch support the switch disables and the explainer swaps to the notice.
- **Dark mode** (`ThemeToggle` rewritten): moon glyph + label + switch row instead of the old icon-button toggle; flips the theme live with the menu open.
- **Legal links & Cookie settings**: Label-5 rows with 16px `fg-brand` glyphs mapped by link name (risk → file-warning, privacy → shield, else file-text; cookie for Cookie settings). Link rows wrap icon+label in a single span because `ExternalLink` HStack-wraps element children, which swallowed the flex gap.
- Icon size note: the Figma MCP codegen for the comp claims 12px icons, but the node metadata has 16×16 frames and the exported SVG assets are 16-viewBox lucide paths at stroke 1.333 — lucide at `size={16}` reproduces them exactly.
- The comp's **Upgrade DAI/MKR row is omitted**: the upgrade surface is parked (`/convert/upgrade` redirects to `/convert`, E2), so the shortcut stays out until that product decision changes.

Rendered values audited against the node metadata + variable defs (panel 274/20/24, row pitch 36, icon→label gap 8, Label 5 at 14/16 −0.28px, Body 7 at 11/16, switch 28×16 with 12px thumb) via computed styles in Chrome.

### Testing steps:

1. `pnpm dev:mock`, open the hamburger menu top right.
2. Compare against the DS comp (node 5178-37443): 274px glass panel; Bundle transactions row with switch + explainer + Legal Notice; divider; Dark mode row; iconed legal links; Cookie settings.
3. Toggle Dark mode — theme flips live, switch state tracks it, panel stays open. Both themes render the glyphs in brand purple.
4. Toggle Bundle transactions — persists via user config (same behavior as the old topbar toggle). With a connected wallet that lacks EIP-5792 support, the switch disables and the unsupported notice appears.
5. Cookie settings closes the menu and reopens the consent banner (PostHog-gated, unchanged behavior).
6. `pnpm exec vitest run` in `apps/webapp` — suite green (1918 tests + 7 new on the two rewritten components).

### DIFF
--- a/apps/webapp/src/components/BatchTransactionsToggle.test.tsx
+++ b/apps/webapp/src/components/BatchTransactionsToggle.test.tsx
@@ -1,0 +1,74 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { I18nWidgetProvider } from '@/widgets/context/I18nWidgetProvider';
+import { BATCH_TX_LEGAL_NOTICE_URL } from '@/lib/constants';
+import { BatchTransactionsToggle } from './BatchTransactionsToggle';
+
+const mocks = vi.hoisted(() => ({
+  batchEnabled: false,
+  setBatchEnabled: vi.fn(),
+  isConnected: false,
+  batchSupported: true as boolean | undefined
+}));
+
+vi.mock('@/modules/ui/hooks/useBatchToggle', () => ({
+  useBatchToggle: () => [mocks.batchEnabled, mocks.setBatchEnabled] as const
+}));
+
+vi.mock('@/hooks', async importOriginal => {
+  const actual = await importOriginal<typeof import('@/hooks')>();
+  return { ...actual, useIsBatchSupported: () => ({ data: mocks.batchSupported }) };
+});
+
+vi.mock('wagmi', async importOriginal => {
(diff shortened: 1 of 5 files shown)


inclinedmind (approved):
Clean, low-risk redesign matching the DS comp. Verified Switch API usage, useConnection hook, and the ExternalLink single-child workaround. Fixed the stale ThemeToggle glyph-size comment (16px). LGTM.

inclinedmind (approved):

PR #1747, opened 2026-07-21
Optimize Envio history queries: bounded, 2-document fan-out with keyset pagination (APP-397)

codewiththeflowxyz:
Optimizes every Envio-backed history query in the webapp. The query shapes predated the subgraph → Envio migration: unbounded result sets, inefficient filter operators, and a fan-out of ~17 parallel requests for the all-networks views, refired on every window focus. Indexer telemetry attributed the large majority of query execution time to these shapes.

Closes APP-397.

## What changed

**Phase 1 — bound and tidy (`c5d2b97`)**
- Normalized address filters to exact-match on lowercased values in all history queries (stored case verified per entity)
- `order_by: { blockTimestamp: desc }` + `limit: 100` on every entity; redundant client-side sorts removed
- 60s `staleTime` on history query keys (history is append-only; previously every refocus refired the full fan-out)
- Trimmed unrendered fields; retired per-contract `useRewardsUserHistory` in favor of the batched hook
- Renamed subgraph-era plumbing (`getSubgraphUrl` → `getIndexerUrl`, etc.)

**Phase 2 — collapse the fan-out & paginate (`31498e9`)**
- Each history hook now exports its query fragments (with optional keyset cursor) and response mapper — single source of truth shared by the per-module hooks and the merged documents
- `useEthereumIndexerHistory`: all mainnet families (savings, upgrade, stake, rewards, stUSDS incl. Curve, sUSDT) in **one document** (was 7 requests); Tenderly dev mode preserved
- `useL2sIndexerHistory`: savings + trade `Swap` aliases for all 4 L2s in **one document** (was 8 requests), with the base/arbitrum PSM cutoffs baked in
- Keyset pagination via `useInfiniteQuery`: cursor on `blockTimestamp: { _lt }`, each page clamped at its completeness boundary (the newest timestamp where an alias hit its limit) so merged rows stay correctly interleaved and pages strictly append
- REST sources (CoW, Morpho, Pendle) merge under the indexer floor so rows never insert mid-list; `BalancesHistory` and the portfolio Transactions table fetch the next page when the user reaches the end
- Net: all-networks history ~15 indexer requests → **2**, all bounded

**Phase 3 — filter-scoped queries (`07f14b5`)**
- The merged documents clamp each page at the densest family's frontier, so filtering to a sparse category (e.g. 30 stake events under 300 swaps) previously showed only its recent rows. The portfolio filters now swap the data source instead of slicing the aggregate: `useHistoryFamilyQuery` fetches one family as its own keyset-paginated document (optionally network-narrowed, spanning the mainnet + L2 indexer legs when the family does), and `useFilteredPortfolioHistory` selects between aggregate, network-scoped, family, and REST sources
- Scoped queries only fire when their filter is selected and share the react-query cache with the always-on sources; filter dropdown options still derive from the aggregate

**Module-table pagination (`a20d200`)**
- All eight per-module history hooks (mainnet/L2 savings, stake, upgrade, rewards, stUSDS, sUSDT, PSM trades) now keyset-paginate through a shared `useHistoryPagination` helper (the merged documents were refactored onto it too), instead of capping at the newest 100 rows per entity
- Every module transactions table fetches the next page when the user reaches the last one — `HistoryTable` (widget tables) gained the same `onPageChange` contract `ProductTransactionsTable` already had
- `useHybridTradeHistory` passes the PSM side's pagination through under the CoW feed (every CoW trade post-dates every PSM trade); `useL2CombinedHistory` replaces its inert pagination stub with a real two-source floor merge, which `useFilteredPortfolioHistory` now passes through
- Morpho, Pendle, and the CoW order feed remain intentionally one-shot REST lists (server-bounded, not Envio)
- `CustomPagination` fix (`b67ad70`): the button window is now derived from `(currentPage, totalPages)` each render instead of stored in click-handler state — the stored layout went stale (current page dropped from the window; page numbers beyond the real total) whenever the page count grew mid-interaction, which keyset appending now does routinely. Regression-tested for growth, shrink, and the beyond-total drift

**Audit sweep + tests (`18badfc`)**
- Checked the result against [Envio's performance guide](https://docs.envio.dev/docs/HyperIndex/database-performance-optimization); normalized three more exact-match filters outside the history hooks (`useTotalUserStaked`, `useStakePosition`, `useUserDelegates`). Remaining substring matchers are genuine search boxes
- Tests across the new layer: clamp-and-refetch pagination cycles for both documents, cross-source floor merging and limiting-side advancement, `useFilteredPortfolioHistory` source-selection matrix and CoW cutoff/floor merging, PSM fragment shape (cutoff + cursor merged into one comparison object), the table's `onPageChange` contract, and the view's `rows`/`optionRows` split

## Measurements

| Metric | Before | After |
| --- | --- | --- |
| Indexer requests, all-networks portfolio feed | ~15–17 per load, refired on every window focus | **2** (+≤2 only while a product/network filter is active); 60s `staleTime` |
| Query bounds | none — full per-entity history every request | every list: `limit: 100` + `order_by: blockTimestamp desc` + keyset `_lt` cursor |
| Response size per page (test wallet, gzipped) | unbounded | mainnet doc ~1.9 KB, L2 doc ~8.2 KB |

(Full before/after query-cost measurements are in the Linear ticket.)

## Verification
- Typecheck clean, ESLint 0 errors, prettier clean
- Unit suite 2057/2057; vnet-backed hooks suite green (538 tests; one pre-existing flaky vnet balance assertion in `openLockFreeSky` passes in isolation/clean runs)
- Every merged document and family document live-validated with the real fragment builders, cursor variants included
- New tests: per-module pagination cycle (stake, the densest document), L2 combined floor merge + limiting-side advancement, hybrid trade pass-through, `HistoryTable` page-change contract
- Browser audit (2026-07-21, fresh Tenderly fork, mock wallet): portfolio load issues exactly the 2 merged documents (bodies inspected — 24 mainnet lists / 12 L2 aliases, all bounded and ordered); selecting the Savings product fires exactly 2 additional family-scoped requests and the table repaginates at the family's own density (the sparse-category fix end-to-end); network filter, last-page navigation, explorer links and status pills behave; console free of errors/React warnings. One pre-existing dev-only quirk (not from this PR): Pendle rows carry real-mainnet `chainId: 1`, absent from the mock chain list, so their Network label renders blank in dev; production unaffected

## Notes / follow-ups
- Curve stUSDS history lost its "entity may not exist" graceful degradation (now part of the merged document); schema presence verified on both the production and Tenderly endpoints
- Indexer-side optimizations are tracked in a separate follow-up ticket
- When a portfolio filter narrows results to under one page, older matches load only as pages advance; the filter-scoped sources make those pages advance at the category's own density


inclinedmind:
Great work, thanks for looking into this. Three items that claude found:

**1. Leftover `_ilike` the audit sweep missed (please fix in this PR)**
`useHasSealEnginePosition.ts` still queries `SealUrn(where: { owner: { _ilike: "${address}" } })`. The PR description says the remaining `_ilike`s are genuine search boxes — this one is an exact address match and fires for every connected wallet, so it's still the un-indexable scan shape this PR exists to eliminate. Assuming `SealUrn.owner` is stored lowercase like the other entities, it's a one-liner: `owner: { _eq: "${address.toLowerCase()}" }`. Worth the same live probe as the others to confirm stored case.

**2. Question: was the pagination mid-range window change intentional?**
Production's `CustomPagination` shows a run of three adjacent page numbers mid-list (`[4] [5] [6] … [20]`); the rewrite collapses that to `[1] … [5] … [20]`. The stateless derivation (which is great, and clearly necessary once `totalPages` grows mid-interaction) doesn't require this — `[1, '…', current − 1, current, current + 1, '…', totalPages]` is equally stateless and stays within the old 5-button budget. There's no design spec on either side (the DS has no Pagination component and the Transactions comps don't render one), so: if the leaner window was deliberate, say so and it's settled; if it was incidental, I'd restore the neighbors to preserve production muscle memory.

**3. Pre-existing, out of scope — filed as APP-401 so it doesn't get lost:**
`HistoryTable` and `BalancesHistory` slice with a raw, unclamped `startIndex`. If the row set shrinks mid-pagination (e.g. wallet switch re-keys the hook, smaller history arrives, no remount), the stale offset slices past the end → `rowsToShow` is empty → `shouldShowError` renders **"No transactions yet" while data exists**, and since the error state replaces the pagination there's no control left to recover with. `ProductTransactionsTable` already defends against exactly this (`paginate.ts` clamps out-of-range pages — C4/APP-327); the same clamp in these two closes it. Not caused by this PR — appends keep `startIndex` valid — but this PR touches both files' pagination paths, so it's cheap to fix here if you'd rather.

Also: the indexer-side follow-up referenced in APP-397 (`@index` additions, counter entities) is now filed as APP-400 and added to the backlog — sectioned so the index additions can ship ahead of the counter entities if the deploys end up decoupled.

codewiththeflowxyz:
Thanks for the careful pass — all three addressed in 757aa7267.

1. **Fixed** — and the sweep had missed a second identical one: `StakingUrn.owner` in `useStakeUserPositions`. Both entities live-probed (owners stored lowercase, `_eq` on the lowercased address returns identical rows), both switched to `_eq`. Re-verified in the browser afterwards: the actual request bodies now carry `_eq` and return the correct data. The remaining `_ilike`s are the two delegate search boxes.

2. **Incidental, not deliberate** — restored your stateless form; mid-list is `[1] … [c−1] [c] [c+1] … [N]` again (regression tests updated). Doing so surfaced that there are *two* `CustomPagination` copies and the earlier rewrite only covered `modules/ui`: the widgets copy still had the stateful window. Its only current consumer (`BalancesHistory` in the wallet drawer) runs in infinite-scroll mode, so the paginated path isn't reachable in today's UI, but since the drawer's row set now keyset-appends I moved it onto the same derived implementation anyway — both copies now share one `pageWindow` helper.

3. **Took the cheap fix here** — `HistoryTable` slices through `paginate()` (same clamp as `ProductTransactionsTable`) and `BalancesHistory` clamps its offset, with a regression test for the shrink-to-"No transactions yet" symptom. Thanks for filing APP-401/APP-400.

Browser-verified end to end with the seeded test wallet (limit temporarily at 5): last-page landings grow the page count live (6 → 13) with the window re-deriving correctly around the current page, and the drawer's infinite scroll advances across keyset fetch boundaries. Console clean.

inclinedmind (approved):

PR #1758, opened 2026-07-24
Drop stale transaction-engine callbacks after session teardown (APP-416 blocked-state bug)

iceanddust:
Fixes the last open code item on APP-416: accepting a transaction in the wallet and closing the modal in the same instant left the app showing only a "Transaction in progress" toast on every subsequent supply attempt, until a page reload.

## Root cause

An in-flight write outlives its host. Dismissing the modal during `INITIALIZED` tears the session down (`handleClose`: status → IDLE, `activeConfig`/`configRef` → null), but if the wallet had already accepted, the orphaned engine's `onStart` fires *after* that teardown. The shared `txCallbacks` applied it unconditionally, stranding `txStatusRef = LOADING` with no modal left to restore — and the one-transaction-at-a-time guard in `launch()` then blocked every new flow forever.

The same race has a second face: after an abandon-then-relaunch, the *old* engine's late `onSuccess`/`onError` landed on the **new** session — advancing its status screen and firing the new config's `onSuccess` for a foreign transaction.

## Fix

- A session generation, advanced by `launch()` and `handleClose()`. Each engine callback closes over the generation of the render that created it and drops itself when the generation has moved on (same pattern as the existing `sessionId` gate on `updateModalContent`).
- Defense in depth: the `launch()` in-progress guard now only blocks when there's a live config to restore — if state ever strands again, the next launch self-heals instead of bricking.

## Tests

- Two regressions in `transactionAbandon.test.tsx` (late `onStart` after close → next launch opens normally, no block toast; late `onSuccess` after abandon-relaunch → replacing session untouched, its `onSuccess` not fired). Both verified red against the unfixed provider.
- The context-suite harnesses now capture callbacks post-launch — the way real engines do (they read the live context value when the mutation starts) — the pre-launch capture was an artifact that generation binding surfaced.
- Full `modules/ui/context` suite 51/51; typecheck/eslint/prettier clean; webapp unit suite 2085/2086 (the one failure is the pre-existing date-rotted `EarnFeaturedCards` Pendle fixture, red on clean `app-redesign` too).

Touches the dismissal-semantics code that landed via #1754 — @codewiththeflowxyz may want a look.


inclinedmind:
> [!NOTE]
> This comment was written by Claude.

One gap: the generation guard only protects engines whose host unmounts on teardown. It works by closure staleness — a late callback is only dropped if the engine is still holding an old render's callback.

That holds for the `backgroundContent`-hosted flows (savings, stUSDS, rewards, vault, claim, upgrade, pendle supply/withdraw): close unmounts the host, its closures freeze at the pre-bump generation, and late fires are dropped.

But the review-first flows (`useConvertLaunch`, `usePendleRedeemModal`, the stake takeovers) keep their engine host mounted on the page across teardown. Those hosts re-render after a relaunch, and react-query invokes the **latest** render's callbacks: `MutationObserver.setOptions` runs every render and pushes the new options into the live `Mutation`, which calls `this.options.onSuccess` — and the receipt-side `useEffect` in `useWriteContractFlow` fires with current-render closures too. So a late callback from these engines carries the current generation and passes the guard.

I verified this by re-running the new regression tests reshaped for a mounted host — reading `txCallbacks` from the live context value at fire time instead of a pre-close snapshot. The abandon-relaunch case fails on this branch: the old session's late `onSuccess` fires the **new** config's `onSuccess` for the foreign tx. (The close-then-late-`onStart` case passes even in this shape — the `configRef` self-heal covers it — so the blocked-state bug itself is fixed everywhere; it's the relaunch corruption that survives, and only for the review-first flows.) Happy to share the test file.

inclinedmind (approved):

PR #1759, opened 2026-07-27
APP-419: render the portfolio callout from first paint, skeleton only the heading while the TVL loads

codewiththeflowxyz:
## Summary

app419-callout-skeleton-desktop
app419-callout-skeleton-mobile

The top-of-page onboarding callout on /portfolio (`SavingsTvlCallout` / `AllocateStablecoinsBanner`) waits for both the earn-marketplace and stablecoin-balances queries to settle before rendering, so it popped in late and pushed the earnings card (and everything below) down abruptly.

The slot now renders the simulate callout from first paint. Everything in it is static copy except the TVL figure, so the card shows its real subtitle and button immediately; while the TVL query is in flight (`tvlUsd: undefined`) the heading wears a skeleton dress — the real copy kept in the layout as transparent text over per-line pills (`box-decoration-clone` gives each wrapped line its own), with an invisible $0.00b stand-in for the figure. Because the skeleton *is* the settled layout, its height matches at every width and locale by construction, and the pills simply dissolve into the headline when the figure lands.

## Verification

- Measured in-browser with the mock wallet and the `/overall/` fetch artificially delayed: callout height and the earnings card's Y position are identical before/after settle — 256px / Y=529 at 390px, 140px / Y=370 at 1440px. Zero layout shift.
- `typecheck`, `eslint`, `prettier` clean; all 89 portfolio unit tests pass. Locale catalogs are build-generated (gitignored), and the heading string had no translations yet, so the move to Lingui component interpolation costs nothing.

## Known limitations (noted in a component comment)

- The slot optimistically shows the simulate callout while the gate settles: users settling into the taller allocate banner, or into no callout at all, still see a shift/swap — the outcome can't be predicted while loading.

Closes APP-419


inclinedmind (approved):

iceanddust (commented):
Adding the skeleton I think makes sense, however this change has the side effect in the opposite scenario: when the banner should not be displayed, it is now displayed for a while and then disappears. Not sure what the optimal solution is here, maybe a more general page skeleton? Or maybe we can ask Properly 

https://github.com/user-attachments/assets/98ace3fb-c9fe-4627-a963-11bb81efb9df

PR #1761, opened 2026-07-27
APP-418: network fee estimation across the transaction modals

iceanddust:
Closes [APP-418](https://linear.app/ekliptyka/issue/APP-418/add-network-fee-estimation-hook-wire-into-transaction-modals)

The app had no gas estimation — every "Network fee" row rendered `NO_VALUE` (`–`). This adds `useNetworkFee` and wires it into the first screen of all nine transaction modals.

The bundled-vs-sequential **savings UI is deliberately deferred**. The hook returns both figures plus the saving ratio, so it can land later with no rework.

## How it estimates

Two concurrent `eth_simulateV1` calls:

- **Sequential** — the N calls as separate entries in one simulation, summed. The EIP-2929 access list resets between entries (verified: three identical `balanceOf` calls in one block each cost 28,867), so this is exactly what signing them one at a time costs.
- **Bundled** — one self-call with Multicall3's runtime code placed at the user's address via `stateOverrides`, modelling the single intrinsic charge and warm storage a 7702 delegate gets. Undelegated accounts add the 12,500 authorization tuple.

Issued as two separate concurrent requests, never two blocks of one request — blocks inside `blockStateCalls` carry state forward, so each would prime storage slots for the other.

Also covers the OP-stack L1 data fee on Base/Optimism/Unichain, excluded on Arbitrum where it is already baked into the returned gas units.

Measured on mainnet (approve + deposit into sUSDS): **199,222 sequential vs 162,592 bundled — an 18.4% saving**, or 12.1% including the first-bundle auth tuple.

## Three bugs found by testing on real mainnet

**1. The fee was 15x too low.** A real batch ([0xb2d7c987…](https://etherscan.io/tx/0xb2d7c9873703fc24762516d0004ff499c4652b37b61d30b484db3e10c54d0005)) was shown as $0.03 and cost $0.59. The gas units were fine (150,612 actual vs our ~162k); the whole error was the price. viem's `estimateFeesPerGas` prefers `eth_maxPriorityFeePerGas`, which reports the tip needed to be included *at all* — the Sky proxy returned 0.000426 gwei. Wallets bid for prompt inclusion instead: MetaMask paid a 2.0 gwei tip against a 0.1 gwei base fee, and when the base fee is that low the tip **is** the price.

Now derived from `eth_feeHistory`: next block's base fee plus the median p95 priority fee over 20 blocks. Against that transaction's own block, p95 reproduces what it paid to within 0.01% (2.100810 vs 2.100674 gwei); p90 reaches only 69% and p99 overshoots to 106%.

**2. The same transaction priced differently per wallet** — $0.59 on MetaMask, $0.12 on Ambire. Both accounts are 7702-delegated, so both took an ERC-7821 path against their own delegate. MetaMask's answers correctly (148,070 gas); **Ambire's returns _success_ from an `execute()` it does not implement**, reporting 31,180 gas for work costing ~148,000 — and we trusted it, because we only checked the call didn't revert.

The real-delegate path is gone. The stand-in executor matches MetaMask's delegate to 0.05% (148,003 vs 148,070), so consulting the wallet bought nothing while risking a 5x error. Both accounts now price identically. A batch cheaper than its own most expensive call is additionally discarded rather than displayed.

**3. The row quoted a fee before an amount was entered**, pricing a zero-value flow. Estimation now waits for the same flag that gates the confirm button.

## Which gas price we bill at — resolved

Wallet tip policies differ ~15x, so the percentile choice is really "which wallet tier do we mirror". Measured on the same blocks:

| | price | our figure | wallet quoted |
|---|---|---|---|
| base fee only | 0.101 gwei | $0.028 | MetaMask Low $0.04 |
| **+ p50 tip** | 0.201 gwei | **$0.056** | **Ambire Slow/Medium $0.06 · Fast $0.07** |
| + p95 tip | 2.101 gwei | $0.581 | MetaMask Market $0.59 |

We bill at **p50** — the gas-weighted median tip actually paid — so the row states what Ethereum charges rather than what one vendor bids. MetaMask floors its tip at 2 gwei against a ~0.1 gwei base fee (~10x the going rate) and pays it in full; priority fees are never refunded, only unused `maxFeePerGas` headroom is. 13% of transactions in recent blocks paid under 0.01 gwei in tip and were still included, so that floor buys little in this regime.

A tooltip on the fee row owns the part we cannot know:

> Estimated at the current network rate: the base fee plus a typical priority fee. Your wallet chooses the final priority fee, so your cost may be higher or lower.

## Plumbing

To reach the calls without touching ~20 engine hooks, `useTransactionFlow` and `useWriteContractFlow` now return the calls they will send plus whether those go out bundled; everything that spreads their result inherits it. Each launch hook passes that through to its modal, which keeps its existing dash as the fallback.

Nine surfaces, first screen only: savings, vault, stUSDS, rewards, convert, claim, stake claim, pendle, upgrade.

## Open items

- **Which gas price to bill at is still an open product decision.** Wallet policies differ ~15x: p95 matches MetaMask's Market default, while the median transaction on the network pays ~$0.04 and Ambire's tiers quote $0.06–0.07 for this flow. **p50 is what's currently implemented** (`PRIORITY_FEE_PERCENTILE = 50`) — the market rate, per the table above; p95 would track MetaMask's Market tier and overstate for users on efficient wallets. Options discussed: a range, the market rate, or the wallet-default end.
- **The stake claim modal draws one fee row but offers two CTAs** (Claim / Claim & Restake). It prices the plain claim, the action always available; restake's extra cost is not represented. Flagged in code.
- **Design**: the savings comp implies ~90% bundling savings; reality is 12–18%. Relevant when the savings UI is built.

## Testing

- Unit + hooks suites green (583/583 hooks). The one failing unit test, `EarnFeaturedCards`, fails identically on `app-redesign` — pre-existing.
- **This feature cannot be verified on a Tenderly fork**: vnets' `eth_simulateV1` ignores the vnet's own mutated state (`eth_call` balanceOf → 49,999 USDS, `eth_simulateV1` → 0), so flows relying on injected balances revert in simulation and the row shows `–`. Fork base fee is also 1 wei. Verified against real mainnet instead. Worth knowing before anyone writes an e2e assertion on that row.


inclinedmind (commented):
Great estimation core — well-documented, tests pin the real failures. Before merge:

**1. Fee tooltip (`NetworkFeeLabel.tsx`):** DS `Tooltip` is hover-only and force-closed on touch, so the explanation is unreachable on mobile. `BundleTogglePanel` already uses a popover for exactly this reason — use the `InfoTooltip` hybrid (hover desktop / popover touch, same chrome).

**2. `useConvertLaunch` memo deps:** the memo captures `bundleState`/`networkFee` but deps miss `ready`, `canBundle`, `isBatch`, `sequentialFormatted`. If capabilities resolve after the fee lands, the badge never appears (`NetworkFeeValue` can't re-render itself). Convert-only — other surfaces are unmemoized.

**3. Stale fee on cleared amount:** `keepPreviousData` still surfaces the old result once `enabled` goes false — clear the input, row keeps the last fee. Return `undefined` when `!amountReady`.

**4. Copy:** "bundled transactions saves" (grammar, ×2 — could be one shared component); "Approve + Supply" shows on claim modals, where the bundle is N claims.

**5. Minor:**
- PR body still says p95; code bills p50. Screenshots missing.
- `useBundleFeeState` re-derives `canBundle` — expose from `NetworkFeeData` instead.
- No tests for the badge/promo visibility matrix.
- With `BatchToggle` removed, the in-modal toggle needs the badge → gone if estimation fails. Nav toggle covers it; QA note.

1 and 2 are the ones I'd hold for.

iceanddust:
Thanks — all of it was reproducible. Five commits, `aac5b8d2a..c21999c37`.

**1. Fee tooltip** (`aac5b8d2a`). Right, and the copy it hides is the part that most wants reading on a phone. `NetworkFeeLabel` composes `InfoTooltip` now; same content, same 14px icon.

**2. `useConvertLaunch` deps** (`ec09e84ae`). Confirmed, and worse for someone who already had bundling on: `promoVisible` is permanently `false` for them, so it couldn't force the recompute either. Now lists `isBatch`, `sequentialFormatted`, `settled` and `canBundle` as primitives — depending on the objects would defeat the memo and re-open the update loop it guards.

**3. Stale fee** (`e5401314d`). `placeholderData` applies whenever the observer has no data of its own, and a disabled query has none. Returns `undefined` when `enabled` is false.

**4. Copy** (`c21999c37`). One `BundleExplainer` for both surfaces, which retires the grammar slip. "Approve + Supply" was wrong on more than the claim modals — convert is Approve + Convert, upgrade likewise, and both claim modals bundle N claims with no approval at all — so the shared copy names the steps generically ("The steps of this transaction run as one"). The comps do say Approve + Supply; per-flow copy would have to come from each flow's `steps` and I'd rather have a design answer than guess. Flagging that as the one open thread here.

**5.** PR body fixed — it billed p50 in the table and p95 in Open Items; `PRIORITY_FEE_PERCENTILE = 50`. Badge/promo matrix now has tests (14, in `NetworkFeeValue.test.tsx`).

The one I'd rank higher than you did is the last bullet: with `BatchToggle` gone the badge is the only bundling control on the screen, and gating it on the fee having landed means a call that reverts in simulation takes the switch away with the number — and on a Tenderly fork it is never there at all. So it's a QA note *and* a real hole. `BundleFeeState` grows `settled` (the estimate is in **or** has failed) beside `ready` (the figure is in): the badge reads `settled`, the value and the savings card still read `ready`, so the row keeps changing shape once. Each surface passes the fee query's error through (`dfe044852`).

The one I've left alone is `useBundleFeeState` re-deriving `canBundle`. Exposing it from `NetworkFeeData` would tie a wallet-capability fact to a simulation succeeding, which is the coupling the paragraph above just removed — and `useIsBatchSupported` is a shared query, so the duplication costs nothing.

Verified on a fresh mainnet fork, dark mode, batch mock wallet. The fork is the failed-estimate case by construction: fee reads `–` and the `Bundled` badge sits beside it, the panel opens without dismissing the modal, the tooltip opens on hover and — with `ontouchstart` present — on tap, and clearing the input drops the row back to a plain dash with nothing held over. Gates green (2107/2108; `EarnFeaturedCards` fails identically on `app-redesign`).

inclinedmind (approved):

PR #1763, opened 2026-07-28
APP-426: redesign small issues/bugs pt. 2

inclinedmind:
For APP-426.

- Desktop header top offsets aligned with the comps (Convert/Portfolio/Stake were 40px below the nav; comps pin 80/80/96)
- Portfolio title now steps down to 24/26 below `md` per the mobile comp (the one title missing the ramp)
- Portfolio statistics heading copy → "Sky Protocol Statistics"
- Transactions table value column → "Amounts" (per ticket item 5; diverges from the comp's "Supplied")
- Reward claim rows resolve the paid token (SPK/GROVE/…) from the reward-contract registry instead of defaulting to USDS
- Fixed Yield rows resolve the market's underlying (sUSDS) for the icon instead of falling back to avatar text
- Borrow slider drag/typing lag fixed (item 3b): the slider and risk readouts derive from a live vault simulation on the raw amounts — the per-amount work is pure math over cached chain reads — while validation errors and the RPC-bound calldata prep stay on the debounced amounts
- Borrow input's top percent chip is now 75% instead of 100% (the exact max sits on the on-chain liquidation boundary, so fee accrual alone tips a maxed position underwater); repay keeps its 100% chip with wipeAll staging
- Pendle history amounts show the PT ticker again ("0.21 PT-sUSDS") in portfolio transactions and the market-detail history table — APP-399's rename of the market config name to the product display name had leaked into the unit label

Item 3a (the mid-track max marker) is deferred to APP-433 pending design input; the far-left slider disappearance is logged there too.


codewiththeflowxyz:
The claim icon in light mode is white and is barely visible. Dark mode works fine though.
image

codewiththeflowxyz (approved):
Left a small comment about the claim rewards icon in the transactions table, the rest look good to me.

iceanddust (approved):

PR #1768, opened 2026-07-28
Fix audit gate: bump axios/postcss overrides, prune expired overrides

inclinedmind:
Unblocks the failing \`pnpm audit --prod --audit-level high\` gate (and the dev→staging promotion PR). Two new high advisories published against transitive deps:

- **axios** [GHSA-gcfj-64vw-6mp9](https://github.com/advisories/GHSA-gcfj-64vw-6mp9) (`>=1.15.2 <1.18.0`) — our own override was pinning into the vulnerable range; bumped `^1.16.0` → `^1.18.0`
- **postcss** [GHSA-r28c-9q8g-f849](https://github.com/advisories/GHSA-r28c-9q8g-f849) (`<=8.5.17`) — new override → `^8.5.21` resolved

Both are Node-only/build-time code paths, so no runtime exposure for the shipped webapp — this is CI hygiene.

Also pruned 30 self-expired overrides: regenerated the lockfile with `overrides: {}` and checked every selector against natural resolutions — only `esbuild@<0.28.1`, `@json-rpc-tools/provider>axios@^0.21.0`, and `ws@>=8.0.0 <8.21.0` still match anything, so only those remain. Audit after: **0 high** (moderates also dropped 23 → 14 from fresher transitive resolutions).

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -285,39 +285,10 @@ catalogs:
 
 overrides:
   esbuild@<0.28.1: ~0.28.1
-  '@babel/runtime@7.26.7': 7.26.10
-  axios@>=1.0.0 <1.16.0: ^1.16.0
-  '@json-rpc-tools/provider>axios@^0.21.0': ^1.16.0
-  protobufjs@<=7.6.0: ~7.6.1
-  '@protobufjs/utf8@<1.1.1': ~1.1.1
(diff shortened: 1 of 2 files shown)


inclinedmind:
Superseded by #1769 — same axios/postcss fix without the override pruning, which we're deferring to a ticket for more careful testing.

PR #1770, opened 2026-07-28
Portfolio: available rewards sections (Ecosystem + Merkl) + claim modal restyle

iceanddust:
Builds the last unimplemented section of the Portfolio comp: the two claimable-rewards tables between the position cards and Transactions ([Figma 1036:190212 / 1036:189973](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=386-35313)).

The modals already existed — D5's `ClaimRewardsPanel` — but not in a form these sections could launch, so the work split three ways.

## What's here

**1. New claim scopes.** `ClaimScope` covered "everything", one Morpho vault, one sky reward contract and one staking urn — nothing addressed *every Merkl token* or *this one Merkl token*, which is exactly what a section's Claim all and a row's Claim need. Adds `merkl`, `merkl-token` and `sky-rewards`. Staking rewards stay off the Portfolio by construction: the stake adapter yields nothing for any of these.

Two supporting additions to the normalized reward model let the rows read straight off the adapters rather than through a second, parallel read: `tokenName` (the full name under the symbol — it can't come from the reads, since the Merkl API returns a symbol only and `TOKENS.name` mostly repeats the symbol, so the names live in a map like `idleView`'s) and `ClaimableResult.refresh`, so a surface can settle its rows after a claim. `sourceLabel`/`badge` are dropped — they only fed the modal's group headers.

**2. The claim modal restyled to its comp** ([1036:190105](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1036-190105) / [1036:190315](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1036-190315)): amount heroes (32px logo, Heading-2 amount, inline `($78.90)`, token pill) over a two-column Network fee / Network footer. **The comp shows no checkboxes, even in the 2-token variant** — the launching control now decides what gets claimed, so the selection state goes and everything in scope is always claimed. Gating, the merged `Call[]` and the single `useTransactionFlow` are untouched.

This lands on **every** consumer of the shared panel — the vault position card and the /earn/rewards farms — which is the intent: one claim modal, one look. Sky `getReward` farms also pick up the comp's "Claim ecosystem rewards" title.

**3. The sections themselves.** Ecosystem rewards (SPK / GROVE / CLE, plus the deprecated SKY one while a balance remains), then Merkl Rewards. The table is the comp's `Table / Idle` instance, so it reuses the idle stablecoins table's cells verbatim and its below-md `TransactionCard` fallback (no reward-specific mobile comp exists). Both self-hide when empty.

## Decisions worth a look

- **CTA hierarchy is yours, not the comp's** — the comp only ever drew a one-row ecosystem section and a two-row Merkl one. More than one reward → primary on Claim all, row buttons secondary; a single reward → no Claim all, row button promoted. Since that flips as the read settles, the loading state renders a same-shape skeleton and the heading row is pinned to the button height, so neither transition moves the page.
- **Two comp details not followed**, both reading as leftovers from the copied idle table: the ecosystem row's button says "Supply" (it opens a claim modal), and its Claim all is hidden rather than absent.
- **Section order** is Ecosystem → Merkl per your call; the two comps disagree (each puts its highlighted section first).
- **Transactions heading** bumped 18px → 24px. The comp puts all three section titles on Heading 5; left alone Transactions sat visibly smaller than the two new sections above it. Separate commit — easy to drop.
- `Network fee` still renders `–`. Same TODO as the Savings/Vault modals; APP-418 is the fix.

## Verification

Gates: typecheck clean, lint 0 errors, prettier clean, 2093/2094 unit tests. The one failure — `EarnFeaturedCards` "18 Jun 2026" — fails identically on `app-redesign` and is unrelated.

Browser-verified on a Tenderly fork in dark mode, with real on-chain ecosystem rewards (restarted SPK + deprecated-SKY farm emissions, staked, warped):

- both sections render per comp, in order, between the cards and Transactions
- **both CTA variants side by side** — Merkl 2 rows (Claim all + secondary rows), Ecosystem 1 row (no Claim all, primary row)
- row Claim opens the single-reward modal, claim-all opens the multi-row one — no checkboxes in either
- claimed SPK for real: tx succeeded, the row dropped out, Claim all disappeared and the surviving row's button flipped to primary, live (so `refresh` is wired correctly)
- no "Maximum update depth" loop — the D5 failure mode this panel had a history of

Merkl rows were exercised against a stubbed Merkl API response (the fork has no off-chain Merkl state); the on-chain ecosystem path was entirely real.

Not covered: e2e (per branch convention), and GROVE/CLE rows specifically — GROVE isn't deployed at the fork block.


iceanddust:
Pushed a fix for the "click too fast → no transaction" report (3869a05).

**Cause.** `useSequentialTransactionFlow`'s `execute` needs a simulated request; without one it hits its `not ready to execute` branch, logs, and returns. The panel gated on each *adapter's* calldata readiness but never on the *flow's* own `prepared`, so confirming early walked the modal to the wallet screen having dispatched nothing. The panel mounts as `backgroundContent`, so its `useTransactionFlow` starts cold on **every** modal open — which is why it's so easy to hit right after a page load.

**Fix.** One term added to the same gate the savings and vault bodies have always used:

```ts
const disabled = calls.length === 0 || preparing || !flow.prepared;
```

**Coverage — this fixes the product pages too**, since they all launch this one panel:

| Surface | Path | Status |
|---|---|---|
| Portfolio reward tables | `openClaim({kind:'merkl-token'/'reward-contract'/…})` | fixed |
| `/earn/rewards` farm cards | `RewardsPositionCard` → `openClaim({kind:'reward-contract'})` | fixed |
| Morpho vault position card | `VaultPositionCard` → `openClaim({kind:'vault'})` | fixed |
| Stake page | own `useStakeClaimLaunch` | no change needed — `StakeClaimModal` already gates on `plainPrepared`/`restakePrepared` before launching |

**Verified in the browser.** The window is sub-frame against a local fork, so I throttled simulate/estimate RPCs by 3s to widen it (which is also the real-world condition where this bites) and sampled the CTA's disabled state from the moment the modal opens:

- Portfolio row Claim → CTA appears **disabled at 684ms**, enables at **4670ms** when the simulation lands
- `/earn/rewards` SPK card → **disabled at 559ms**, enables at **4544ms**

Pre-fix the button was live that whole window and a click there silently did nothing.

Also covered by a unit test asserting confirm stays disabled while the flow reports `prepared: false`. Gates re-run: typecheck clean, prettier clean, 2094/2095 (the one failure is the pre-existing `EarnFeaturedCards` one, unrelated).

iceanddust:
Filed [APP-435](https://linear.app/ekliptyka/issue/APP-435/wire-network-fee-estimation-into-the-restyled-claim-rewards-modal) for the `Network fee` row (Todo, blocked by APP-418).

Heads up while scoping it: **this branch and #1761 collide in `ClaimRewardsPanel.tsx`.** #1761 wires the fee into the panel's *old* stacked `InfoRow` footer and adds a `BundleSavingsPromo` under it — but this branch replaced that footer with the comp's two-column `FooterStat` block and deleted `InfoRow`, and its copy of the panel still assumes the per-reward checkbox state removed here. `ClaimRewardsPanel.test.tsx` is edited by both too.

Since #1761 is In Review and this is still a draft, the likely resolution is rebasing this branch onto it and absorbing the fee wiring during that rebase — in which case APP-435 closes with the rebase rather than as separate work. Decision recorded there: `BundleSavingsPromo` renders full-width beneath the two columns, as the claim comp has no slot for it.

inclinedmind (commented):
Nice work, one potential issue and a couple small comments:

Opening any claim modal render-loops (silently — the tab stays usable at 100% CPU, so fork testing won't show it). useMerklRewards's mutate gets a new identity every render (array-literal queryKey in its deps); wired into useClaimable's memo it makes rewards fresh every render → fresh transactionScreenContent → useModalEntryBody effect → updateModalContent → re-render → repeat. Verified in vitest: mutate identity fails across rerenders; a churning body in the real provider loops until OOM.

Fix in useMerklRewards.ts:
const queryKey = useMemo(() => ['merkl-rewards-all', userAddress, chainId], [userAddress, chainId]);

Nits: 
1. hero USD needs minDecimals: 2 (($78.9) vs comp's ($78.90)); 
2. Refresh deps can be [ecosystem.refresh, merkl.refresh] once mutate is stable.

iceanddust:
Thanks @inclinedmind — all three were valid, all three are in (branch also merged up to the latest `app-redesign`).

**1. The render loop (1ae2e32e1).** Confirmed, and it's worse than Merkl-only: because `useMerklClaimable`'s memo lists `mutate`, it recomputes every render and returns a **fresh `rewards` array even when empty** — so *every* scope loops, not just the Merkl ones. Applied your `useMemo` fix; verified it's sufficient (with it, the panel's merged `[...merkl.rewards, ...sky.rewards]` is referentially stable across rerenders). The sky adapter is clean — `refetch` is stable — and `stakeAdapter` already carried this stabilization explicitly.

Worth writing down for the next adapter, since the trap is in the shared plumbing rather than in Merkl: `updateModalContent` does `setActiveConfig(prev => ({ ...prev, ...rest }))`, so state *always* changes, and the provider's `value={{ … }}` is an inline literal — every push re-renders every context consumer, panel included. Anything churning that reaches `useModalEntryBody`'s sync effect closes the circle. And since the setState is in a passive effect, React never throws "Maximum update depth" — it just spins, exactly as you described.

Added `merklAdapter.identity.test.tsx` as a regression guard (`exhaustive-deps` can't catch a fresh array literal in a dep list). It fails on the pre-fix code and passes after. One harness gotcha for anyone extending it: the `QueryClient` has to be hoisted outside the wrapper — a per-render `new QueryClient()` churns `useQueryClient` and the test then fails even *with* the fix.

**2. Hero USD (0b91ccf24).** Fixed with `{ minDecimals: 2, maxDecimals: 2 }`. Also an internal inconsistency, not just a comp mismatch — the row that launches the modal uses `formatUsd`, which pads, so the same value read `$78.90` in the table and `$78.9` in the modal. (`formatUsd` itself isn't reusable here; the hero renders its own `$`.) The other hero call sites, e.g. `StUsdsAmountSummary`, don't pad either — left that alone as out of scope.

**3. Refresh deps (e3aa8ca7a).** Done, with one deviation: `[ecosystem.refresh, merkl.refresh]` trips `react-hooks/exhaustive-deps` (it can't see through the property access and asks for the whole objects back), so I destructured `refreshEcosystem` / `refreshMerkl` first — same effect, no new warning.

Gates: typecheck clean, prettier clean, lint 0 errors / 252 warnings (one fewer than before, none in the touched files), units 2120 passed. The single failure is `EarnFeaturedCards.test.tsx` "18 Jun 2026", a date-bomb that fails identically on the pristine branch.

inclinedmind (approved):

PR #1773, opened 2026-07-29
APP-432/APP-437: Transaction-modal redesign — all modules migrated onto the shared modal primitives

codewiththeflowxyz:
# APP-432 — Transaction-modal redesign (QA round July 28)

Migrates the transaction modals to the redesigned DS chrome, one module per commit — all modules are complete (see the checklist at the bottom); remaining follow-ups are tracked in Linear (APP-440, APP-441), not in this PR.

Every module follows the same loop: Figma excavation (geometry from `get_metadata`, exact paints from the Plugin API) → token-mapped implementation on the shared primitives → computed-style delta loop in the browser until zero deltas (±1px) → 50%-opacity overlay + side-by-side composites → **live end-to-end transaction on a funded Tenderly fork** → light-mode and mobile passes. In the side-by-side images below, **Figma is always on the left, the app on the right**.

Engines, calldata, and analytics are untouched in all three modules — these are presentation-layer migrations.

## Shared primitives (grew with each module)

- `components/product/ModalAmountField` — DS Input/Amount: label, 24px token icon, Heading-3 input, balance line, 25/50/100% chips, selector slot, hairline.
- `components/product/ModalSummaryGrid` + `CardField*` — the two-column labelled grid with centered hairlines (32px dividers on entry, 24px on review).
- `components/product/ModalGridCells` — semantic cell hints shared by all modules: `token`, `network`, `trend`, `trailingToken`, `rateAccent: 'savings' | 'morpho'` (green-% gradient vs boosted-stars glyph), `productIcon: 'default' | 'morpho' | 'pendle'` (ringed product icons), `labelBadge` (e.g. slippage Auto/Custom pill), `action` (inline interactive element, e.g. the slippage gear).
- Rows are pure, unit-tested data builders per module (`savingsModalRows` / `vaultModalRows` / `pendleModalRows`) — the Figma grid contract (labels, pairing, order) is pinned in CI.

---

## 1. Savings (`5350383`)

Supply/withdraw restyled to the new 610px card: DS amount field, summary grid ([Rate | Network], [Supply | Est. earnings (1Y)] deltas, [Network fee]), three-screen flow (entry → "Review supply" → wallet) with the amount hero, and the green rate-% gradient accent.

| Entry (Figma vs app) | Review (Figma vs app) |
|---|---|
| savings entry | savings review |

<details>
<summary>Live transaction (wallet steps) + light theme</summary>

savings wallet
savings light

</details>

## 2. Morpho vaults (`a1beae6`)

Vault supply/withdraw on the same recipe. Vault-specific treatments: the **boosted-rate stars glyph** (morpho gradient `#60A9FF→#2E8EFF`, drawn only when the market carries reward incentives — validated live on the USDS Flagship vault), the **morpho-gradient product ring** on review, and the static single-asset pill (Figma draws a dropdown, but a vault has exactly one underlying — divergence kept deliberately, no dead chevron). This module extracted the shared `ModalGridCells` so later modules become row-builders + config.

| Entry (Figma vs app) | Review (Figma vs app) |
|---|---|
| morpho entry | morpho review |

<details>
<summary>Boosted stars (zoom), live success, light + mobile passes (vs the Savings representative comps — Morpho has no own light/mobile comps)</summary>

stars zoom
morpho success
morpho light
morpho mobile

</details>

## 3. Pendle PT-sUSDS (`131c856`)

The deepest module: real multi-token dropdown (28px DS pill on Radix Select — supply sUSDS/USDS/USDC, withdraw receive-token), rate **delta** with the green-% accent on both values (market implied → trade effective APY), maturity economics ([Supply | Est. earnings (ND)] and [You'll claim | Claim date] from the PT-at-maturity model: claim = PT balance, supply = its present value at the implied rate), and the **slippage control moved from the modal header into the review grid** — label + Auto/Custom badge + value + inline 14px gear opening the existing SlippageMenu popover. Both flows executed live end-to-end (supply 10,000 USDS → 10,166.89 PT; withdraw 25% of the position).

| Entry (Figma vs app) | Review (Figma vs app) |
|---|---|
| pendle entry | pendle review |

Light and mobile verified against Pendle's **own** comps (1030:59547 / 1222:18959):

| Light (Figma vs app) | Mobile (Figma vs app) |
|---|---|
| pendle light | pendle mobile |

<details>
<summary>Withdraw review, slippage popover from the grid gear, live success</summary>

pendle withdraw review
pendle slippage
pendle success

</details>

## 4. Rewards/Claim (`93cb7b8`)

The generalized claim modal (`modules/claim`, launched from the Morpho vault card and the rewards position cards — this also closes the vault claim modal deferred from the Morpho section). The body becomes **hero reward rows** — 32px token icon, Heading-2 amount with the USD value inline in parens, and a 28px token badge pill (extracted from `TransactionAmountHero` as the shared `TokenBadge`) — over the standard [Network fee | Network] grid. Per the QA-round comps the previous per-token **checkbox list and source group headers are gone**: a claim launch claims its full scope. Engines untouched (same adapter trio merged into one batched flow); the stake-scope restake toggle survives. Verified live end-to-end on a funded fork: supplied 5,000 USDS to the SPK farm, time-warped a week, claimed 61.37 SPK through entry → wallet → success. Light + mobile verified by token symmetry / sheet reflow (no own comps).

## 5. Upgrade DAI/MKR (`2293a04`)

The Upgrade modal (More-menu launch, APP-413) restyled onto the shared primitives — engine and calldata untouched. Entry is **single-screen** (the comps draw no review stage; "Continue" goes straight to the wallet screen): DS amount field with the shared **`TokenSelectorPill`** (extracted from Pendle) as the DAI/MKR source dropdown, over the summary grid [Rate | You'll receive], [Penalty ⓘ | Network] (penalty is MKR-only), [Network fee]. Two new shared grid hints ship with it: the **token-pair rate cell** (`◉ 1.00 = ◉ 24,000.00`) and **`labelAction`** (the 12px penalty info popover after the label). The confirm screen gets the comp's hero: source → target rows (32px icon, Heading-2 amount, token badge pill) with the 20px move-down arrow. Verified live end-to-end on a funded fork: 7,500 MKR → 172,800,000 SKY through Approve + Upgrade (the 4% delayed-upgrade penalty applied on-chain and previewed in the grid). Light + mobile verified by token symmetry / sheet reflow (no own comps).

65-sbs-upgrade-entry-figma-vs-app

66-sbs-upgrade-confirm-figma-vs-app

## 6. Convert/Trade (`9affc57`)

The Convert review modal (review-first flow: the swap form lives on the page, so the modal opens straight on "Review conversion") restyled onto the shared primitives — PSM engine and calldata untouched. The review body becomes the **shared `TokenTransferHero`** (extracted from the Upgrade confirm screen into `components/product`, now reused by both modules): from → to rows (32px icon, Heading-2 amount, token badge pill) joined by the 20px move-down arrow, over the summary grid [Rate `◉ 1.00 = ◉ 1.00` | Network], [Slippage | Fee], [Network fee]. The wallet screen swaps its old one-line text summary for the same hero, and the approve step renders the "Approve ◉ USDS" token chip; per the comp it keeps the title "Review conversion" (no "Confirm in the wallet" variant). Verified live end-to-end on a funded fork: 10,000 USDS → 10,000 USDC through Approve + PSM convert, on-chain verified. Light verified against its own comp (1030:59493); mobile sheet reflow verified.

75-sbs-convert-review-figma-vs-app

79-sbs-convert-confirm-figma-vs-app

## 7. Stake claim (`39227bc`)

The stake claim modal (position details → "Claim rewards") migrated onto the shared TransactionModal. The old flow was three dialogs deep — a hand-rolled entry `Dialog` with a per-token checkbox list, then the shared modal's review screen, then the wallet screen; the comps collapse it to **entry → wallet**: a shared-modal entry ("Claim rewards": reward hero rows + [Network fee | Network] grid) whose CTAs execute directly into the **"Confirm claim"** wallet screen ("Claim amount" relabelled hero above the `Claim ◉ SKY → Restake ◉ SKY` steps). The comps draw no per-token selection, so the checkbox machinery is gone — the flow claims the urn's full claimable set, same call as the generalized claim modal. New shared chrome capability: an optional **secondary confirm CTA on entry-only flows** (`TransactionEntry.secondaryConfirmLabel` + `onSecondaryConfirm`) drawing the comp's two flex-1 buttons — secondary "Claim" beside primary "Claim & Restake SKY" — with each CTA routed to its own engine (plain claim = adapter calls; restake = the F1 calldata seam, both untouched). Click-time `steps`/`analytics` pushes now land on the provider's config ref synchronously so the engine's `onMutate` tracks the clicked mode, retries re-fire the last-clicked engine, and live pushes freeze once the tx leaves the entry (the post-claim refetch would otherwise blank the executed heroes — caught in the browser loop, unit-pinned). `TransactionAmountHero` gained `size="lg"` (40px icon + Body-5 USD line, this module's hero variant) and stake claim amounts pin two decimals. Verified live end-to-end on a funded fork, both paths: **Claim & Restake** (22.9 SKY → vat ink +22.90, earned zeroed) and **plain Claim** (wallet +22.90 exactly, ink unchanged). Light + mobile sheet verified (no dedicated claim comps there — token symmetry).

## 8. stUSDS (`d0287ac`)

The stUSDS supply/withdraw modals (Tether Savings, APP-266 family) restyled onto the vault-family template — **no dedicated comps exist**, so the grids follow the approved spec on the Savings/Morpho recipe with the provider-routing surfaces kept intact: the route notice (reason/warning states), the ≥2% price-impact acknowledgement, and the one-time expert-risk acknowledgement, all restyled to DS tokens with logic and copy untouched. Entry: DS amount field (static USDS pill, 25/50/100% chips wired to a new `setPercentAmount` that preserves the whole-share max flag) over [Rate | Network], [Supply | Est. earnings (1Y)] USDS-delta cells, [Network fee]. Review: from→to `TokenTransferHero` (supply USDS→stUSDS with the quoted receive amount — the Curve route's price impact makes the quote material; withdraw stUSDS→USDS) over the review grid with a **Route cell** (label badge Native/Curve + "stUSDS module"/"Curve pool" value — the Pendle slippage-badge precedent). Rate stays plain (no incentive accent) and Product draws the neutral ring. Rows pinned by a 23-test builder suite.

<details>
<summary>Empty entry, supply wallet/success, withdraw entry + review</summary>

</details>

## 9. QA-round shared fixes (`eb14560`, `7fc65a9`, `d26d709`, `ea95758`)

Four cross-module fixes from this round's verification sweep:

- **Mobile amount field** (`eb14560`) — `ModalAmountField` now reflows to the Figma mobile arrangement (selector pill beside the amount, percent chips full-width) instead of flex-wrapping the desktop grouping, and the stray blue focus ring on the sheet container on open is gone. Closes the shared-chrome mobile TODO.
- **Savings withdraw review** (`7fc65a9`) — withdraw now has its own review stage per comp 859:36322: "Review withdrawal" title and withdraw-specific rows. Closes the Savings follow-up TODO.
- **Live-push freeze gate** (`d26d709`, sweep finding B-4) — `useModalEntryBody` freezes its live config pushes once the tx leaves IDLE, so mid-flight allowance/balance refetches can no longer collapse the executed step list or blank the review breakdown on the wallet/status/failure screens. The gate Stake claim and Convert already had, hoisted so every module inherits it; unit-pinned.
- **Batch toggle card** (`ea95758`, sweep finding C-2) — the legacy "Bundle transactions" hairline row + info popover becomes the comp's "Save on clicks & network fee" card (1036:206739 off / 1036:207086 on): bordered 16px-radius card, the exact DS `lightning-filled` glyph (new `modules/icons/LightningFilled`, fg-brand) beside a Label-4 header and the M-size DS switch, with a Body-6 explainer that names the flow's own steps ("Approve + Supply run as one transaction…" on Savings, "Approve + Convert…" on Convert, etc.).

### Disclosed divergences (all deliberate, consistent across modules)

- Entry CTA reads **"Review"** (some comps draw the action verb) — matches the three-screen flow shipped since Savings.
- The amount input shows the raw typed value while editing (comps draw a formatted filled state).
- Percent chips are 28px vs the comps' 26px, and the card's top padding is 4px tighter than the comps — both shared-chrome constants inherited from the approved Savings chrome, identical in every module.
- Pendle: withdraw-review comp omits the slippage row; it's slotted in as [Withdrawal | Slippage], [Network | Network fee] because the sell quote uses slippage identically. Titles/product follow the PT naming convention ("PT-sUSDS"), not the market's marketing name.
- Rewards/Claim: the comps' `/Confirm` frames draw an in-place loading CTA, but the Stake-claim comp shows the claim family on the shared wallet/status screen — the standard flow is kept for machinery parity (minimize/toast, retry, success). Old Morpho-section claim comps (859:38365+) are drift; the Portfolio set wins.
- Upgrade: the DAI entry comp draws a Penalty row, but the delayed-upgrade penalty only exists on MKR→SKY — it's omitted for DAI (showing "4.00%" on a penalty-free upgrade would be wrong data). Step 2 reads "Upgrade MKR to SKY" as plain text where the comp inlines two token icons (the shared steps model renders one trailing token chip). The MKR comp's trigger pill drawing "USDS" and its "1.20" rate are placeholder drift — the real 1:24,000 contract rate is shown.
- Convert/Trade: the convert step reads "Convert USDS to USDC" as plain text (same two-inline-icon comp limitation as Upgrade). Network fee is stubbed "–" pending the cross-module gas-estimate work. The rate pair's inner icon gap keeps the shared 4px (this comp draws 3px, the Upgrade comp 4px — designer micro-drift).
- Stake claim: the confirm side-by-side is captured at the success state (the mock wallet auto-signs too fast to hold the comp's "Confirm in the wallet" moment — same screen, later lifecycle). An "Approve ◉ SKY" step prepends when the SKY allowance is short (the comp draws the allowance-covered two-step case). The app's SKY token icon draws the DS pinwheel on a white disc while the DS `crypto-logos/sky-full` uses a navy `#1A173F` disc — an app-wide asset drift visible in every module, tracked as a global follow-up below.
- Wallet-screen back arrow: the Savings-era DS wallet comp (859:36215) draws back + title + close and the shared chrome follows it in every module; the newer Portfolio-section confirm comps (Convert 1036:205564, Stake claim 1036:214007) draw title + close only — comp-vs-comp drift, the established chrome is kept (the arrow is also the error-path route back to the two-CTA entry).
- stUSDS: no dedicated comps — vault-family template throughout. The supply review swaps Est. earnings for the receive quote (Curve price impact makes it material; earnings already showed on entry); the Route cell's badge shape is the Pendle slippage Auto/Custom precedent.
- Batch toggle card: the comp's bare "Learn more" text is wired to the batch-transactions legal notice (the only in-app explainer — same target the nav-menu toggle discloses). The ON comp's "Bundled" badge + strikethrough price in the Network fee cell belongs to the separate network-fee PR (fees are stubbed "–" across all modules here).

## TODO

- [x] **Rewards/Claim** modals (1036:190079+/190318+) — done, incl. the vault claim modal deferred from the Morpho iteration (shared `modules/claim`); the "Claim ecosystem rewards" title variant ships with the future portfolio ecosystem surface
- [x] **stUSDS** — done on the vault-family template per the approved spec (no own comps; designer FYI outstanding)
- [x] **Upgrade DAI/MKR** (1310:130712+/130658+) — done; restyle of the APP-413 modal onto the shared primitives
- [x] **Convert/Trade** (1036:205509/205564) — done; review body + wallet screen on the shared primitives, `TokenTransferHero` extracted for reuse
- [x] **Stake claim** (1036:213978→214007) — done; entry + wallet screen on the shared primitives, chrome gained the secondary-confirm entry CTA (takeover modals are a separate ticket)
- [x] **Failure/bundled states** re-verified per migrated module (1036:207845+/206742+) — swept via three parallel verification sessions; blockers triaged: the shared engine hang is pre-existing on `development` (APP-438, own PR), the bundled false-success is a mock-connector env gap (APP-439), and the two branch regressions (B-4 freeze gate, C-2 batch toggle card) are fixed here (section 9)
- [x] **Shared chrome, mobile** — done (`eb14560`): amount-field mobile reflow + sheet focus-ring fix
- [x] Savings follow-up: withdraw-review title vs comp 859:36322 — done (`7fc65a9`)

All migration items are done. The remaining follow-ups are not part of this PR and are tracked in Linear: the sweep's cosmetic backlog + open designer questions (APP-440) and the `tokens/sky.svg` white-vs-navy disc asset drift (APP-441, reference images attached there).

---

Screenshots are served from the `app432-pr-assets` orphan branch (commit-SHA-pinned raw URLs); it can be deleted once this PR is merged and the images are no longer needed.

https://claude.ai/code/session_017jb25X8h6AiQ5y564d8kiR


apps/webapp/src/modules/pendle/components/pendleModalRows.ts
```diff
@@ -0,0 +1,208 @@
...
+ * [Slippage | Network], then Network fee full-width. Withdraw follows
+ * 859:41679 with the Slippage cell slotted in ([Withdrawal | Slippage],
+ * [Network | Network fee]) — the comp omits slippage, but the sell quote uses
+ * it the same way the buy does, so the control must stay reachable.
+ */
+export function buildPendleReviewRows(
+  flow: 'supply' | 'withdraw',
+  input: PendleReviewRowInput
+): PendleModalGridRow[] {
+  const slippageCell: ModalGridCell = {
```
iceanddust, line 151: **Price impact is no longer shown anywhere in the Pendle flow.** The old entry rendered a `Price impact` row from `quote.priceImpact`; neither the new entry grid nor either review grid has it, and slippage only appears on review.

I can't find it in the comps either, so this may well be an oversight on the design side rather than a deliberate cut — worth checking with them. For an AMM swap it's material risk information, and the withdraw review grid has a free cell if it needs to go back in.

iceanddust (commented):
Reviewed the full diff. The module-per-commit structure, the pure row-builder split and the in-code rationale for every Figma divergence make a 4k-line diff genuinely reviewable — nice work. Engines/calldata are untouched as claimed, and I verified the new synchronous `configRef.current` assignment in `updateModalContent` is consistent with every other `setActiveConfig` site in the provider.

A few things worth a second look before merge, plus some smaller cleanup suggestions inline.

apps/webapp/src/modules/pendle/components/pendleModalRows.ts
```diff
@@ -0,0 +1,208 @@
...
+    label: 'Network',
+    value: input.network,
+    network: true
+  };
+  const feeCell: ModalGridCell = { kind: 'single', label: 'Network fee', value: input.networkFee };
+
+  if (flow === 'supply') {
+    return [
+      [
+        { kind: 'single', label: "You'll claim", value: input.claimAfter, token: input.displaySymbol },
```
iceanddust, line 176: **"You'll claim" on the review is the _total_ position at maturity, not what this transaction buys.**

On the entry it's a `before → after` delta so it reads correctly, but here it collapses to `claimAfter` = existing PT + this trade's PT, under a label that reads per-transaction. A user holding 500 PT who buys ~102 sees "You'll claim 602" on the confirm screen.

Worth checking whether this is the desired approach — either the label wants to say it's the total ("Total at maturity"), or the review should keep the delta the entry draws.

apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
```diff
@@ -223,9 +234,15 @@ export function useStUsdsTransactionForm({
     [isSupply, amountLabel]
   );
 
+  // The from→to hero the review and wallet screens draw: supply is USDS →
+  // quoted stUSDS; a withdraw redeems/swaps stUSDS (the quote's input) → USDS.
+  // Scalar deps keep the memo stable (matches the savings/vault forms).
+  const quotedStUsds = isSupply
```
iceanddust, line 240: `selectedQuote` is `StUsdsQuote | undefined` (loading, refetching, all-blocked), so the `?? 0n` fallback means `TokenTransferHero` can render a fabricated `100.00 USDS → 0.00 stUSDS` on the review/wallet screens rather than a "no quote yet" state.

Preferably pass the `–` placeholder through (or gate the review CTA on a live quote) so a transient zero never reads as a real quote.

apps/webapp/src/modules/ui/components/TransactionModal.tsx
```diff
@@ -397,16 +454,40 @@ export function TransactionModal({
...
-                <Button
-                  variant="primary"
-                  size="xl"
-                  className="w-full"
-                  onClick={handleConfirm}
-                  disabled={firstScreenConfirmDisabled}
-                >
-                  {firstScreenConfirmLabel ?? <Trans>Confirm</Trans>}
-                </Button>
+                {showBatchToggle && <BatchToggle steps={steps ?? []} />}
```
iceanddust: `showBatchToggle` is gated on `isFirstScreen`, which in the new three-screen flow is **both** the entry and the review — so the user walks past the "Save on clicks & network fee" card twice in a row.

It seems a bit redundant to show it on both screens. It makes more sense to show it only on the first screen, i.e. the entry and not the review — something like `(!hasReviewStage || isEntry)` alongside the existing condition.

apps/webapp/src/modules/ui/components/TransactionModal.tsx
```diff
@@ -473,47 +554,46 @@ export function TransactionModal({
...
+        />
       </div>
-      <Switch
-        checked={batchEnabled}
-        onCheckedChange={setBatchEnabled}
-        aria-label={t`Toggle bundled transactions`}
-      />
+      <Text className="font-graphik text-fgSecondary text-xs leading-[18px]">
+        <Trans>
+          {stepLabels} run as one transaction. You sign once and bundled transactions saves you clicks and
```
iceanddust: Nit: `stepLabels` is a pre-joined, already-localized string interpolated into the middle of a `<Trans>` sentence. It extracts fine, but the sentence is only grammatical in English-like word order — a translator can't move the subject. Not blocking while the catalogs are empty, just flagging the shape.

apps/webapp/src/components/product/ModalGridCells.tsx
```diff
@@ -0,0 +1,256 @@
...
+    </>
+  );
+}
+
+/**
+ * 12px stars glyph in the morpho gradient (Figma Icons/Custom/stars-filled,
+ * gradient-morpho #60a9ff→#2e8eff top→bottom) — the boosted-rate marker. Glyph
+ * geometry matches the widgets `Sparkles` icon; only the fill differs.
+ */
+function MorphoStars() {
```
iceanddust, line 78: Nit: `MorphoStars`, `GradientRing` and `TrendIcon` are inline SVG here, while this same PR correctly adds `LightningFilled` to `modules/icons`. Inconsistent with the convention the PR otherwise reinforces — these would sit more naturally alongside it.

apps/webapp/src/components/product/ModalGridCells.tsx
```diff
@@ -0,0 +1,256 @@
...
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
+}
+
+/** 12px chain icon for the Network cells. */
+function NetworkIcon() {
+  const chainId = useChainId();
```
iceanddust, line 117: Nit: `NetworkIcon` reads the connected `useChainId()`, and the callers derive `networkName` from the same id — but Pendle's engine runs on `engineChainId` (mainnet/tenderly), not necessarily the connected chain. Same value in practice today; might be worth threading the engine chain in rather than relying on that holding.

apps/webapp/src/components/product/ModalGridCells.tsx
```diff
@@ -0,0 +1,256 @@
...
+      </span>
+    );
+  }
+  return (
+    <span className="flex flex-wrap items-center gap-1.5">
+      <span className="flex items-center gap-1">
+        {icon}
+        <span>{accent(cell.before)}</span>
+      </span>
+      <ArrowRight className="text-fgPrimary size-3 shrink-0" aria-hidden />
```
iceanddust, line 231: Nit: the delta branch renders the same `icon` element in both halves. Harmless today (only token deltas exist), but a cell with `network: true` + `kind: 'delta'` — which the type allows — would draw the chain icon twice.

apps/webapp/src/components/product/TokenSelectorPill.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const face = (symbol: string) => (
+    <span className="flex items-center gap-1">
+      <TokenIcon token={{ symbol }} width={16} showChainIcon={false} className="size-4" />
+      <span className="font-circle text-fgPrimary text-sm leading-4 font-medium tracking-[-0.28px]">
+        {symbol}
+      </span>
+    </span>
+  );
+
+  if (tokens.length <= 1) {
```
iceanddust, line 35: Nit: this static-chip branch is duplicated twice more as module-local `AssetPill`s (`VaultModalForm`, `StUsdsModalForm`) with the same markup and class string. Those two could just render `TokenSelectorPill` with a single-element list and drop their local copies.

apps/webapp/src/components/product/ModalSummaryGrid.tsx
```diff
@@ -0,0 +1,47 @@
...
+  className
+}: {
+  rows: ModalSummaryCell[][];
+  /** Vertical-hairline override — the entry screen draws 32px (`h-8`), the review 24px (`h-6`). */
+  dividerClassName?: string;
+  className?: string;
+}) {
+  return (
+    <div className={cn('flex w-full flex-col gap-4', className)}>
+      {rows.map((row, i) => (
```
iceanddust, line 30: Nit: array indices as keys. Fine for the static grids, slightly risky where the row shape changes between renders — e.g. the savings builder re-pairs its last row when `minReceived` appears on L2.

apps/webapp/src/modules/savings/components/savingsModalRows.ts
```diff
@@ -1,110 +1,153 @@
...
   earningsBefore: string;
   /** 1Y estimated earnings after — stubbed until a projection source exists. */
   earningsAfter: string;
-  /** Network the transaction runs on (e.g. "Ethereum"). */
-  network: string;
   /** Network fee, formatted — stubbed until a gas estimate is wired. */
   networkFee: string;
 };
 
+const singleOrDelta = (
```
iceanddust, line 47: Nit: `singleOrDelta` is duplicated verbatim in four row-builder files (savings, vault, pendle, stUSDS), each carrying the same `as ModalGridCell` casts. Worth hoisting next to the type it constructs in `ModalGridCells`; a small generic on the `base` param would let the casts go too.

apps/webapp/src/modules/savings/components/SavingsModalForm.tsx
```diff
@@ -116,80 +98,123 @@ export function SavingsModalForm({
...
+  // draws, over the review grid. "You'll receive": mainnet supply previews sUSDS
+  // shares (convertToShares); L2 supply shows the PSM min-out floor; withdraw is
+  // the requested amount in the destination token.
+  const youReceive = isSupply
+    ? isL2
+      ? `${formatBigInt(engineParams.minAmountOut ?? 0n, { unit: 18, maxDecimals: 2 })} sUSDS`
+      : previewShares !== undefined
+        ? `${formatUsds(previewShares)} sUSDS`
+        : NO_VALUE
+    : `${value ? formatNumber(parseFloat(value), { maxDecimals: 2 }) : '0'} ${originSymbol}`;
```
iceanddust, line 166: Nit: the withdraw branch derives the display value from the raw input string via `parseFloat(value)` rather than the parsed `amount` bigint — which this file's own older comment argued against ("Derived from the parsed engine `amount` (not the raw input) so the preview matches what's submitted"). Not reachable as `NaN` in practice since the review requires `amountReady`, so purely a consistency point with the sibling forms.

iceanddust (approved):

apps/webapp/src/modules/pendle/components/pendleModalRows.ts
codewiththeflowxyz: Restored in #1781 (`90ab32282`). Both review grids now carry a `Price impact` cell next to Slippage — supply pairs `[Slippage | Price impact]`, withdraw `[Price impact | Network]` — fed from `|quote.priceImpact|` exactly like the old modal. Kept it review-only to match where slippage lives in the new flow. The comps indeed don't draw it, so it's flagged in the builder docstring as a disclosed divergence to raise with design.

apps/webapp/src/modules/pendle/components/pendleModalRows.ts
codewiththeflowxyz: Agreed it read as per-transaction — addressed in #1781 (`3b1966b79`). The value stays the total (paired with `Claim date`, "at maturity you claim X on 18 Jun" is the position fact the entry's delta already previews), but the cell is relabeled **"Total at maturity"**, following the comp's own "Total earnings" precedent of saying total explicitly. Disclosed as a label divergence from 859:41264 in the builder docstring.

apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
codewiththeflowxyz: Fixed in #1781 (`25fd04dca`) — the quoted stUSDS leg now stays `undefined` while no quote is live, and the hero renders the `–` placeholder instead of `0.00 stUSDS`. (The review CTA was already gated on `prepared`, which needs a quote to build calls, so the zero could only flash during loading/refetch — now it can't.)

apps/webapp/src/modules/ui/components/TransactionModal.tsx
codewiththeflowxyz: Superseded by the APP-418 network-fee work (`7316b1903`) — the review-screen toggle card is gone entirely. Bundling now lives in the fee cell's Bundled/Not-bundled badge + panel, and the "save on fees" promo card renders only on the entry screen, so there's nothing to walk past twice anymore.

apps/webapp/src/modules/ui/components/TransactionModal.tsx
codewiththeflowxyz: Also gone with the APP-418 commit — that card's copy was replaced by `BundleExplainer`, which deliberately uses a generic, non-interpolated `<Trans>` sentence ("The steps of this transaction run as one…") because the step names vary per flow. No pre-joined string survives.

apps/webapp/src/components/product/ModalGridCells.tsx
codewiththeflowxyz: Moved in #1781 (`906d000a3`): the stars glyph is now `modules/icons/SparklesMorpho` (the DS name `StarsFilled` was already taken by the vault card's *different* sparkle glyph; this one shares geometry with the widgets `Sparkles`, whose static gradient ids it couldn't reuse) and the trend arrow is `modules/icons/TrendingUp`. `GradientRing` stayed local on purpose — it's a ring *wrapper* around children, not an icon, so it doesn't fit the `Icon` signature.

apps/webapp/src/components/product/ModalGridCells.tsx
codewiththeflowxyz: Threaded in #1781 (`d385e9f60`). Grid cells can now carry `networkChainId` (the icon falls back to the connected chain when absent), and the Pendle form passes `engineChainId` for both the Network cell's icon and its name — so the row keeps telling the truth even if Pendle ever becomes reachable from a non-mainnet chain.

apps/webapp/src/components/product/ModalGridCells.tsx
codewiththeflowxyz: Kept as is — for token deltas the double draw is the comp's treatment (`◉ x → ◉ y`), and no builder emits a `network: true` delta; a network delta ("Ethereum → Ethereum"?) would be a builder bug rather than something the renderer should second-guess. Happy to split the hint union so `network` excludes `kind: 'delta'` if it ever bites, but it didn't seem worth the type gymnastics today.

apps/webapp/src/components/product/TokenSelectorPill.tsx
codewiththeflowxyz: Done in #1781 (`9119339b3`) — both module-local `AssetPill`s are deleted; `VaultModalForm` and `StUsdsModalForm` now render `TokenSelectorPill` with a single-element list (which collapses to the static chip). `onSelect` became optional so static uses don't have to pass a noop.

apps/webapp/src/components/product/ModalSummaryGrid.tsx
codewiththeflowxyz: Done in #1781 (`aa0c7d78a`) — rows and cells now key by the cell's `testId` (index only as fallback), so cells keep identity through re-pairs like the savings entry re-rowing when `minReceived` appears on L2.

apps/webapp/src/modules/savings/components/savingsModalRows.ts
codewiththeflowxyz: Hoisted in #1781 (`55d3d0c57`) into `ModalGridCells` next to the cell type — it had actually grown to five copies with the rewards builders. The hints half of the cell type is now its own `ModalGridCellHints`, which types the `base` param directly, so all the `as ModalGridCell` casts are gone with the copies.

apps/webapp/src/modules/savings/components/SavingsModalForm.tsx
codewiththeflowxyz: Fixed in #1781 (`1af11c057`) — the withdraw "You'll receive" now formats the parsed `amount` (`formatUnits` + `formatNumber`), matching the rule the file's own comment states and the sibling forms follow.

PR #1776, opened 2026-07-30
APP-432 item 8: "Products unavailable in the US" section on /earn

iceanddust:
Closes item 8 of [APP-432](https://linear.app/ekliptyka/issue/APP-432/redesign-qa-round-july-28). Figma: [Earn (US regulations) — Table Section 1036:201473](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1036-201473&m=dev).

## What was missing

Every `/earn/*` detail route is already geo-gated (`requireModuleEnabled` → redirect to `/portfolio`), but the marketplace list wasn't: a restricted product still sat in **Earn Opportunities** looking supply-able, and clicking it bounced you to the portfolio. The comp splits those products into a second, dimmed table below the marketplace. The heading node carries a designer annotation making the intent explicit:

> If something is not available, we still show it as the last items.
>
> _(We want to disable the VPN blocker, so if someone has their VPN set to the US, there's a chance they won't see the most important products if we disable them)_

The restricted set is already known client-side — the geo config's per-module `enabled` flags, mapped from each row's `intent`. With the restrictive config that's `savings` + `rewards` + `expert`, i.e. exactly the four rows the comp lists (Sky Savings, Earn SPK, Earn Chronicle Points, stUSDS), leaving vaults + fixed above.

## Changes

- **`EarnPage`** partitions the marketplace rows on the geo config and renders the second table last. Both tables run through the same filters and share one sort, so the two sections read as one list split in two; the section is omitted entirely when the region (or the active filters) leaves nothing to list.
- **`EarnTable`** grows a `dimmed` variant matching 1036:201476 — logos, network/supply stacks and the risk pill at 50% opacity, product name at `fgSecondary`, values at `fgTertiary`, no hover tint, no click/keyboard selection. Its test ids are namespaced (`testIdPrefix`) so two tables can coexist; the default keeps every existing id, including the ones `network-switching.spec.ts` uses.
- **Featured cards** are fed the available rows only, so a restricted product is never given an `Earn` CTA that redirects. (Deviation from the comp — see below.)
- **`Intent` → geo `ModuleId`** moves out of `useWidgetItems` into `geo-config/moduleForIntent`, so the nav filter and the marketplace split share one map. Behaviour-identical: same entries, `upgrade` deliberately still absent.
- New pure helper `partitionByGeoAvailability` + tests, and tests for the dimmed table variant.

While the geo lookup is in flight the provider reports every module disabled, so the split treats everything as available until it settles — otherwise the whole list would flash through the restricted layout on first paint.

## Decisions worth a look

- **Featured cards filtered (deviates from the comp).** The US artboard still features Sky Savings while listing it as unavailable below; its `Earn` button would redirect to `/portfolio`. Confirmed with Hernando to filter instead. Happy to revert to the literal comp if design prefers.
- **Mobile has no comp.** Below `md` the table reflows into accordion cards with Supply / View details buttons. Dimmed cards keep the accordion (the data is still worth reading) and drop both actions. Covered by a unit test; not visually verified — the QA browser wouldn't resize below the desktop breakpoint. Worth a mobile comp from design.
- **Shared sort.** Sorting either table's headers re-sorts both, matching the comp's identical (sortable) header row on the second table.

## Verification

- `pnpm typecheck` ✅ · `pnpm lint` ✅ · webapp unit suite 2132/2133 — the single failure is `EarnFeaturedCards` "18 Jun 2026", which fails identically on `app-redesign` (pre-existing, unrelated).
- Browser-verified on localhost, dark mode: `?geo_mode=restricted` renders the section with the four expected rows dimmed and inert (click does nothing, no hover tint, no NEW badge); `?geo_mode=full` collapses back to the single 11-row table with no section; per-module overrides (`?geo_module_expert=false`) move just that product. Light theme spot-checked.
- No e2e added, per branch convention.


inclinedmind (approved):
Looks good, one potential improvement. Should we add some copy for the edge case where we can't verify the region, something like "We couldn't verify your region, so these products are unavailable."

iceanddust:
I implemented that

PR #1777, opened 2026-07-30
APP-432: wire the 30D Rate column in the earn marketplace table

iceanddust:
APP-432 item 5: the Earn Opportunities table's **30D Rate** column rendered `—` on every row since C1, which left it as the only unwired column in the marketplace. Every product family does expose a daily rate history, so each row now reports a real trailing 30-day average.

## Where each row's history comes from

| Row | Source | Cost |
| --- | --- | --- |
| Savings | BA Labs `/save/ssr/historic/?days_ago=30` | 1 new request |
| Rewards (SPK/GROVE/SKY/CLE) | the farm chart the table already loads | none — the series was already in hand |
| stUSDS | BA Labs `/overall/historic/` `stusds_rate` | shares the detail page's query key |
| Morpho vaults (5) | `historicalState.avgNetApy` | 1 request for all five (see below) |
| Spark sUSDT | the `history` its market payload already carries | none |
| Pendle fixed | `/historical-data?time_frame=day` `impliedApy` | shares the detail chart's query key |

For Morpho, reusing the detail page's historical query would have meant one request per vault, each carrying that vault's whole lifetime of TVL *and* APY. `useMorphoVaultsTrailingRates` instead asks for the net-APY series alone, inside the 30-day window, with one aliased `vaultV2ByAddress` selection per vault — so all five vaults resolve in a single round trip.

## Decisions

- **Pendle rows show the trailing average implied APY.** The data is a real historical series, so the column reflects how the rate on offer has moved. It is not what an existing position earns — a supplied position is locked at its own rate until maturity — so the code says so at the call site. Note the Pendle detail page deliberately has no 30D row (it shows Fixed APY + Underlying APY), so this is the one place the figure appears.
- **Series shorter than 30 days average what exists** rather than dashing until a full window accrues. This is what the product detail pages have always done, so a young vault's row and its detail page agree.
- **Products with no history keep the dash** — points farms (Chronicle) report no rate at all, and a source outage degrades to `—` rather than sinking the row.

## Consolidation

Three surfaces had each grown their own copy of the trailing-average math (the Morpho vault helper, the rewards detail page, the stUSDS detail page). The first commit lifts it into one `trailingAverageRate` helper and points all three at it, so a row and its product page cannot report different numbers for the same product.

## Verification

Gates: typecheck clean, lint 0 errors, 2126/2126 unit tests, and the vnet-backed marketplace suite green with the new assertions.

Browser-verified on a Tenderly fork (dark mode), reading live API data:

| Row | Rate | 30D Rate |
| --- | --- | --- |
| Sky Savings Rate | 3.52% | **3.58%** |
| Earn GROVE | 6.48% | **10.64%** |
| Earn Chronicle Points | – | **–** (correct: points farm) |
| USDT Savings (Morpho) | 3.09% | **3.32%** |
| Fixed Yield (Pendle) | 5.23% | **5.17%** |
| stUSDS | – | **6.78%** |

The USDT Savings vault detail page independently reports **3.32%** for the same vault — the table's slim 30-day window and the detail page's full-lifetime series agree, which is the consolidation doing its job.

**Not positively verified: the Spark sUSDT row.** The staging Spark API currently returns `{"data":[]}` for `/vaults/savings/usdt/historic` and `apy: "0"` for the current figures, so that vault is dormant there — its 30D dashes for the same reason its current rate does. The mapping runs; there is simply nothing to average yet.

Closes APP-432 item 5.


inclinedmind (commented):
Looks good, just one comment:
fetchPendleMarketHistoricalData maps impliedApy: row.impliedApy ?? 0, and those filled zeros now flow into the new 30D average. trailingAverageRate only drops non-finite values, so a few buckets where Pendle's API omits impliedApy silently deflate the shown rate (3 missing days on a 5% market → 4.5%). On the chart a fake 0 at least renders as a visible dip; in an average it's invisible.

 Suggested fix: make impliedApy optional in PendleMarketChartPoint and drop the ?? 0, then flatMap out the undefined points in useEarnMarketplace (same pattern as the stUsdsChart mapping a few lines up). PendleDetailChart needs a small tweak too since it reads point.impliedApy * 100 — and skipping those points there also fixes the false zero-dip on the chart. (One-line alternative — dropping those rows in the fetcher — loses TVL points for days that have tvl but no impliedApy, so I'd avoid it.)

PR #1778, opened 2026-07-30
APP-432 item 3: wire live data into the Earn hero stat badges

iceanddust:
Closes item 3 of APP-432 — "These badges need real, live numbers".

Both Earn hero badges were hardcoded strings (`$11.02B in circulation`, `Operating for 7 years`), with a code comment flagging them as static marketing copy pending a live source.

## What's wired

**Circulation** → `useUsdsDaiData({ limit: 1 })`, the BA Labs `/overall/historic/` total (USDS + DAI). This is the same series `UsdsTotalSupplyCard` and the portfolio totals chart already read, and the endpoint returns newest-first. The mock's `$11.02B` matches this figure, confirming it's the intended source.

**Operating years** → derived from a single `PROTOCOL_START` constant (`protocolStats.ts`), currently **January 2017** → renders "Operating for 9 years".

> ⚠️ The start date is a placeholder pending confirmation. It's one constant, and it drives both the badge and the subtitle's "since 2017" — change it and the two stay in sync by construction.

## Also in scope

The subtitle carried the same two hardcoded numbers, so it's wired to the same source rather than left to contradict the badges over time.

**Rounding:** the subtitle's coarse figure is **floored**, not rounded. The Figma's `$11.02B` → `$11B` hid the problem, but today's real total is `$9.69B`, which *rounds up* to `$10B` — sitting directly under a badge reading `$9.69B` that looks like a bug. Flooring can only ever understate, which reads as deliberate coarseness.

**i18n:** the subtitle is now three sentence-level translation units (they're three independent sentences, so no word-order is broken) which lets the middle one drop out cleanly. Interpolated values are named, so the extracted ids read `{amount} in circulation` rather than `{0}`. The start year is pre-stringified so a grouping locale can't render it "2,017".

**Skeletons:** both live figures get one, sized to the settled content (badge pill at its real 28px/32px height and ~160px width; an inline pill for the subtitle figure) so the hero doesn't reflow when data lands.

**Unavailable state:** the fetch helper swallows failures into an empty array, so "loaded but absent" is the error path too. In that case the badge and the circulation sentence are both omitted rather than showing `$0` or a dangling " in circulation".

## Verification

- `protocolStats.test.ts` — 8 unit tests covering the year arithmetic (anniversary boundary, no negative counts), the uppercase-suffix badge format, and the floor-not-round rule.
- Full suite: **2133 passing**. One failure, `EarnFeaturedCards.test.tsx`, is **pre-existing and unrelated** — it asserts a Pendle maturity of `18 Jun 2026`, now in the past; it fails identically on a clean `app-redesign` checkout.
- Typecheck clean, lint 0 errors, prettier clean, `lingui extract` produces the expected message ids.
- Browser-verified in dark and light: loaded state, loading skeletons, and the data-unavailable state.

### QA caveats

- BA Labs only sends `Access-Control-Allow-Origin` for the production origins, so **the stat cannot load from localhost at all** — the browser blocks it. This is pre-existing and affects every BA Labs surface in local dev, not something this PR introduces. Local verification of the loaded state used a stubbed response with the real payload; **please sanity-check the real number on the Vercel preview.**
- The local browser would not resize past 606px wide, so visual checks were at mobile width. The only `md:` delta here is the skeleton's height (`h-7` → `md:h-8`), mirroring the badge's own existing size recipe — but worth a glance at desktop width on the preview.

### Note on overlap

`feature/app-432-products-unavailable-us` (item 8, PR #1776) also edits `EarnPage.tsx`. The two touch different regions (hero vs. table sections), but whichever lands second may need a trivial rebase.


apps/webapp/src/modules/earn/helpers/protocolStats.ts
```diff
@@ -0,0 +1,56 @@
+import { formatNumber } from '@/utils';
+
+/**
+ * When the protocol started operating, as claimed by the Earn hero (APP-432
+ * item 3). Drives both the "since 2017" of the subtitle and the "Operating for
+ * N years" badge, so the two can never disagree — change this one date and both
+ * follow.
+ */
+export const PROTOCOL_START = new Date(Date.UTC(2017, 0, 1));
```
inclinedmind: should we set this to Dec 18, 2017?

inclinedmind (approved):
Just one question, otherwise looks good!

apps/webapp/src/modules/earn/helpers/protocolStats.ts
iceanddust: Ah good idea, I'll use that

PR #1780, opened 2026-07-30
APP-432 item 21: restyle the stake open-position takeover to the latest comps

iceanddust:
APP-432 item 21 — the "Open a position" stake takeover, revisited against the current comps.

Figma: [Stake SKY / 1](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1036-209505&m=dev) · [Stake SKY / 02 (Borrow & Delegation enabled)](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1036-209698&m=dev) · [Stake SKY / 03](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1036-209632&m=dev)

## What changed

**It's a modal now, not a second page.** The comps layer the takeover over a render of `/earn` — the flow keeps the page visible underneath. `TakeoverShell` drops the opaque `#040434` + app-background repaint for the DS scrim recipe the dialog and sheet overlays already use (`bg-modalOverlay` + `backdrop-blur-[100px]`).

**Chrome and metrics.** Column narrows 660 → 610px, cards sit 12px apart (was 24), 64px of air under the header, and the header takes the comp's 40/20 padding and keeps its hairline on desktop (it was being dropped above `md`). Cards take the Modal/10-12-17 recipe: 32px inset, 32px between header and body.

**Type scale.** Desktop had escalated to Graphik at 14/18px; the comps hold the compact Circular scale at every tier — Label 4 (16/18) headers and delegate names, Body 6 (12/18) labels, Label 5 (14/16) values, Heading 4 (28/30) amounts, Label 6 (12/14) percent chips. Mostly a matter of deleting `md:` overrides.

**Card 1 gains the share-of-balance slider** (Sliders/Standard) and the hairline above it — it stages the same wei as the percent chips, and an amount typed past the balance pins the thumb at 100 rather than running it off the track. The borrow card's amount block gets the same hairline (it had one on phones only).

**Stat dividers now show on desktop**, where they were `md:hidden`. The borrow card keeps its 2×2 grid on phones and becomes one 4-up row from `md` — the middle divider is `hidden md:block`, so `display:none` drops it out of the grid's flow entirely and the mobile centre rule is untouched.

**The three annotated Info icons are real tooltips now** — they were inert. Copy is verbatim from the Figma annotations, with the live liquidation price and min-collateral figures interpolated. `InfoTooltip` carries its own `TooltipProvider` so it works on any surface (same precedent as `RiskTierDetailsTrigger`).

**Risk pill** moves off `orange-400`/`bullish` onto the DS `components/status` palette, matching what APP-432 item 6 did to the risk meters. The "Updated hourly" badge loses its desktop-only check icon — the comp's badge is text-only.

## Decisions taken with Hernando

- Both optional cards stay **off** by default. Frame /1 shows Delegate expanded, but frame /3 is the true empty state and the annotation only deprioritises Borrow.
- The footer row is **inline** (scrolls with the content) as drawn, rather than pinned to the viewport.
- The annotated **micro-animation** on changing stat values is deferred to its own ticket.
- Scope is **open-position only**; item 23 (manage position) stays open. `TakeoverShell` and the amount field are shared, so the manage sheet inherits the new shell chrome and type scale — that's a step toward item 23, not a substitute for it.
- The delegate "Pre-selection?" TBD in the comp is left alone.

## Verification

- `pnpm typecheck`, `pnpm lint` (0 errors), `prettier --check` — clean.
- Unit suite: 2126 tests, 201/202 files pass. The one failure (`EarnFeaturedCards` Pendle maturity date) is a pre-existing date bomb on `app-redesign` — confirmed by running it on an untouched checkout of the base. New coverage for the slider.
- Browser-verified on a Tenderly fork with a funded mock wallet, dark mode first then light: empty state, 25% chip driving the slider, borrow card expanded with all four stats and dividers, all three tooltips, delegate list and selected-row ring, inline footer with Confirm enabled.

## Caveats

- Mobile has no comp for this item. The inline footer applies there too, so Confirm now sits below the fold on phones with a long delegate list — worth a design call.
- Light mode is interim until G2, but it was checked and holds up.

### DIFF
--- a/apps/webapp/src/components/InfoTooltip.tsx
+++ b/apps/webapp/src/components/InfoTooltip.tsx
@@ -1,5 +1,5 @@
 import { Info, X } from 'lucide-react';
-import { Tooltip, TooltipContent, TooltipPortal, TooltipTrigger } from './ui/tooltip';
+import { Tooltip, TooltipContent, TooltipPortal, TooltipProvider, TooltipTrigger } from './ui/tooltip';
 import { Popover, PopoverClose, PopoverContent, PopoverTrigger } from './ui/popover';
 import { useIsTouchDevice } from '@/hooks';
 
@@ -48,17 +48,23 @@ export function InfoTooltip({
       </PopoverContent>
     </Popover>
   ) : (
-    <Tooltip>
-      <TooltipTrigger aria-label="Show additional information">
-        <Info size={iconSize} className={iconClassName} />
-      </TooltipTrigger>
-      <TooltipPortal>
-        <TooltipContent className={contentClassname}>
-          <div className="max-h-[calc(var(--radix-tooltip-content-available-height)-64px)] scrollbar-thin overflow-y-auto">
-            {typeof content === 'string' ? <p>{content}</p> : content}
-          </div>
-        </TooltipContent>
-      </TooltipPortal>
-    </Tooltip>
+    // Own provider (same 300ms delay as the app root's, which nests harmlessly
+    // under it) so the trigger is self-contained on any surface — pages, cards,
+    // full-page takeovers, tests — with no host setup. Same precedent as
+    // RiskTierDetailsTrigger.
+    <TooltipProvider delayDuration={300}>
+      <Tooltip>
+        <TooltipTrigger aria-label="Show additional information">
+          <Info size={iconSize} className={iconClassName} />
+        </TooltipTrigger>
+        <TooltipPortal>
+          <TooltipContent className={contentClassname}>
+            <div className="max-h-[calc(var(--radix-tooltip-content-available-height)-64px)] scrollbar-thin overflow-y-auto">
+              {typeof content === 'string' ? <p>{content}</p> : content}
+            </div>
+          </TooltipContent>
+        </TooltipPortal>
+      </Tooltip>
+    </TooltipProvider>
   );
 }
--- a/apps/webapp/src/components/product/TakeoverShell.tsx
+++ b/apps/webapp/src/components/product/TakeoverShell.tsx
@@ -6,13 +6,13 @@ const FOCUSABLE_SELECTOR =
   'a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])';
 
 /**
- * Full-screen takeover chrome (hi-fi 486:32657): header row with title + badge
- * and a close button, a scrollable centered card column, and a sticky footer.
- * The one sanctioned ProductDetailTemplate exception (Migration Mechanics §5) —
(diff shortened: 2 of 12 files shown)


iceanddust:
Review fixes pushed in 984735df9.

**Applied**
- **Slider percent rounding** (`StakeTakeoverStakeCard.tsx`) — confirmed real. Staging floors and the projection back floored too, so any balance that isn't a round multiple of 100 wei read back `p − 1`; the 25% chip parked the thumb at 24. Projection now rounds. Added a dust-bearing-balance case (`1234567891234567891234` wei) and a keyboard-drag case asserting the staged amount — the old 1000 SKY fixture round-trips exactly, which is why it passed. Re-verified on a fork with a deliberately dusty balance: `aria-valuenow` reads 25.
- **Stat divider alignment** — dropped `md:items-start` on the borrow stats row so its three hairlines centre like card 1's.
- **Confirm button width** — `md:w-40` → `md:min-w-40`, so a longer translated label wraps instead of clipping under the base recipe's `whitespace-nowrap`.
- **Nits** — cited the footer copy's 237px measure; added `aria-valuetext` to the slider thumb. That last one needed a `valueText` prop on the shared `Slider`: Radix puts `role="slider"` on the Thumb, so the attribute can't reach it through Root's props (the first attempt was a silent no-op — verified in the browser that it lands now).

**Not applied**
- **DS `*Bg` token for the risk pill** — fair point, but there are no `statusWarningBg`/`statusErrorBg` tokens, so matching `table-cells.tsx` properly means adding two tokens with light-mode values I'd be guessing at. Leaving the uniform `/10` until someone owns that call.
- **"The comment describes an app root provider that no longer exists"** — this one is incorrect. `pages/App.tsx:52` still has `<TooltipProvider delayDuration={300}>` on this branch *and* on untouched `origin/app-redesign`, and `pages/main.tsx:8` mounts that `App`. The claim that it was deleted appears to come from a grep for `pages/App` that missed main.tsx's relative `'./App'` import. So `InfoTooltip` was not crashing in the running app — only in tests rendered outside the root, which is what the comment says. The observation that the sibling `widgets/shared/.../InfoTooltip.tsx` has no provider of its own is separate and pre-existing.
- **Inline footer reaching `ManagePositionTakeover`** — acknowledged in the PR body; flagged for item 23 rather than changed here.

Gates after the fixes: typecheck, lint (0 errors) and prettier clean; 201/202 test files pass, the one failure being the pre-existing `EarnFeaturedCards` date bomb.

codewiththeflowxyz (approved):

PR #1781, opened 2026-07-31
APP-437: address the PR #1773 review — Pendle price impact + claim label, stUSDS quote placeholder, shared-primitive cleanups

codewiththeflowxyz:
Follow-up to #1773, addressing @iceanddust's review — every inline comment verified against the current code first, then fixed where warranted. One commit per finding.

## Fixes

| # | Finding | Commit |
|---|---|---|
| 1 | **Pendle: Price impact was dropped in the restyle.** Confirmed a regression (the legacy modal rendered `\|quote.priceImpact\|`). Restored to both review grids beside Slippage — supply `[Slippage \| Price impact]`, withdraw `[Price impact \| Network]`. Review-only, matching where the slippage control lives; the comps omit it, disclosed in the builder docstring. | `90ab32282` |
| 2 | **Pendle: review "You'll claim" showed the position total under a per-transaction label.** Kept the total (it pairs with Claim date) and relabeled to **"Total at maturity"**, per the comp's own "Total earnings" precedent. | `3b1966b79` |
| 3 | **stUSDS: `?? 0n` could render a fabricated `100.00 USDS → 0.00 stUSDS` hero.** The quoted leg now stays `undefined` without a live quote and the hero shows `–`; the review CTA was already gated on `prepared`. | `25fd04dca` |
| 7 | **`NetworkIcon` assumed connected chain == engine chain.** Cells can pin `networkChainId`; the Pendle form derives the Network cell's name and icon from `engineChainId`. | `d385e9f60` |
| 6 | Inline SVGs → `modules/icons` (`SparklesMorpho`, `TrendingUp`); `GradientRing` stays local (wrapper, not an icon). | `906d000a3` |
| 9 | Module-local `AssetPill` copies deleted; vault + stUSDS forms use `TokenSelectorPill` (single-element list, `onSelect` now optional). | `9119339b3` |
| 10 | Grid rows/cells keyed by stable `testId` instead of array index. | `aa0c7d78a` |
| 11 | `singleOrDelta` hoisted into `ModalGridCells` (had grown to five copies); new `ModalGridCellHints` type kills the `as` casts. | `55d3d0c57` |
| 12 | Savings withdraw "You'll receive" derived from the parsed `amount`, not `parseFloat(value)`. | `1af11c057` |

**Not changed, answered on the review threads instead:** #4 (batch toggle on both screens) and #5 (`stepLabels` i18n) were already resolved by the APP-418 fee work (`7316b1903` deleted the old toggle card; `BundleExplainer` uses non-interpolated copy) · #8 (delta branch drawing the icon twice) is the comp's token-delta treatment and a network delta is unrepresentable in practice.

## Verification

- Full webapp unit suite green (209 files / 2183 tests), typecheck clean, eslint clean on touched files.
- **Visual A/B against the pre-change baseline** (same fork, wallet, amounts, theme, viewport; per-screenshot pixel diff): savings/vault/stUSDS/rewards entry + review screens are **pixel-identical** — the refactors changed no pixels. The only visual delta is the intended Pendle review change, verified live on a Tenderly fork (real quote: Price impact 0.03%, Total at maturity, live network fee).
- Pendle builder tests updated to pin the new grid shapes; everything else is covered by the existing Figma-contract tests unchanged.


apps/webapp/src/modules/pendle/components/PendleModalForm.tsx
```diff
@@ -412,6 +415,8 @@ export function PendleModalForm({
     : NO_VALUE;
   const slippageDisplay = `${formatNumber(slippage * 100, { maxDecimals: 2 })}%`;
   const slippageMode = slippage === defaultSlippage ? t`Auto` : t`Custom`;
+  // |impact| like the legacy modal — direction doesn't matter for risk display.
+  const priceImpactDisplay = quote ? formatDecimalPercentage(Math.abs(quote.priceImpact)) : NO_VALUE;
```
iceanddust: The legacy modal didn't take the absolute value — it negated, deliberately, with the reason spelled out (`widgets/PendleWidget/components/SupplyWithdraw.tsx:128-132` pre-deletion):

```ts
// Pendle's API uses positive = favorable; we display with the inverse
// convention so positive = unfavorable (matching TradeWidget/StUSDSWidget
// and broader DeFi conventions). The raw quote.priceImpact stays unflipped
// for analytics/debugging.
const displayPriceImpact = quote?.priceImpact !== undefined ? -quote.priceImpact : undefined;
const priceImpactRow = displayPriceImpact !== undefined ? `${(displayPriceImpact * 100).toFixed(3)}%` : '—';
```

So with `Math.abs`:

- **A favorable quote (raw `priceImpact > 0`) now renders as a cost.** Legacy showed `-0.02%`; this shows `0.02%`, which under the app-wide convention reads as slippage against the user. Same magnitude, inverted meaning.
- Precision drops 3dp → 2dp, so impacts below 0.005% render `0.00%` (legacy: `0.004%`). Defensible for grid consistency, but worth a conscious call rather than a side effect.
- Legacy guarded `!== undefined`. `priceImpact: number` is non-optional in the type, but it's passed straight through from the API (`useQuotePendleConvert.ts:313`) with no runtime validation — unlike the sibling `priceImpactBreakdown`, which *is* type-guarded there. A missing field now yields `NaN%` where legacy fell back to `—`.

Suggest `formatDecimalPercentage(-quote.priceImpact)`, and fixing the comment — it asserts something about the legacy behaviour that isn't the case (the PR description carries the same claim).

Worth noting this is invisible to the tests: the builder test only pins the passthrough string (`'0.02%'`), and nothing covers the derivation here. A two-line unit on the sign would catch it.

iceanddust (approved):
LGTM, just a small thing before merging

PR #1782, opened 2026-07-31
APP-432 item 23: restyle stake manage-position surfaces to the redesign comps

codewiththeflowxyz:
Implements **APP-432 item 23** — "Stake page, manage position: card style and content sizes need to be revisited for the latest Figma styles" — across the two surfaces the comps section covers: the **position-details modal** (comps 1036:214176 debt / 1036:214314 no-debt) and the **manage-position takeover** (1036:213821 + card/filled variants).

## Position details modal
- Glass card (`bg-bgSecondary`, 28px radius) with the manage menu as a **full-bleed darker subsection** (new `--color-modalSubsection` token, dark `#09042033` / light `#bcb6ef4d`) — the shared DialogContent's `sm:px-10/sm:py-8` is neutralized so the panel runs flush to the card edges like the comp.
- Left column: 44/48 Circular heroes, 12px Graphik labels over 14/16 Circular values, vertical hairline seams between stat columns, no more border-boxed grids; the column **distributes vertically** (`justify-between`) when the menu panel drives the card height (the no-debt comp's layout).
- Reward values: white with a **gradient trending-up icon** (new `TrendingUpGradient`, savings gradient) and trailing 12px token icon — replaces the green `+$` treatment.
- Menu rows: 80px pitch, `borderPrimary` hairlines (none after the last row), comps' lucide icons (banknote-arrow-down/up, coins, door-closed), `fgQuaternary` chevrons.
- Claim chip → `Badges/Special` gradient pill: **compact** amount (`123.46M`, not `123,456,789`) + 12px token icon, no symbol text.
- CTAs h-12; the no-debt pair renders side-by-side.

## Manage takeover
- **Summary strip** rebuilt: 16px Circular heading, side-by-side 32px heroes with a 48px hairline, four hugging stat cells with hairlines, white `+$` rewards + token icon, the dash risk badge, and the "Updated hourly" XS badge on `glassBadge`.
- **Cards**: 32px insets, mode pills as Tabs Items (outlined at rest, brand-gradient active), the Balance/Borrowed line moved **above** the percent chips, a full-width hairline under the field, and the stacked info rows became hugging **stat columns** with white→white deltas, token icons instead of "SKY" text, and the status-color risk pill.

### Behavior changes to note
- **Borrow chips are 25/50/100%** per the comps (were 25/50/75). The 100% chip stages the same whole-USDS-floored safe max the slider's full-right stop already reaches.
- **Repay's top-right hint is "Borrowed: N"** per the comp (was "max. N USDS"); the error ladder still guards overdraw.

### Divergences kept (deliberate)
Disabled undesigned rows (Change reward / Close position) keep `opacity-40`; chip gradient angle approximated `to-r`; back chevron kept on the takeover; pill gradient/stroke are raw hexes (theme-invariant, as in the comps); slider min/max labels stay plain text. Inactive/liquidated details states verified by symmetry (no inactive urn on the test fork).

### Verification
Numeric probe (computed styles + rects) hit zero deltas against the comp geometry on both details states, both themes, incl. panel-edge insets = 0; mobile pass clean (mobile comps ship as-is by design). Flows driven live on a Tenderly fork: details → takeover routing, borrow 30k end-to-end (confirm modal → tx → success redirect), claim modal from the chip row. 1045 stake-module tests pass (4 assertions updated, 1 added for the compact chip).

## Screenshots
_Figma | before | after — drag zones below, one image each:_

**Details modal — debt state (1036:214176):**

H-details-debt-figma-before-after

**Details modal — no debt (1036:214314), full-bleed panel + distributed left column:**

I-details-nodebt-figma-before-after

**Manage takeover — summary strip + cards (1036:213821):**

J-manage-takeover-figma-before-after

**Borrow card filled, deltas (1036:213297):**

K-borrow-card-figma-before-after

**Debt state with live claimable chip:**

76-details-debt-with-chip


iceanddust (commented):
### Blocking: the 100% borrow chip stages the exact liquidation boundary

`StakeTakeoverAmountField.tsx:12` replaces this comment:

> *Borrow caps the top chip at 75%: staging the exact max puts the position on the liquidation boundary, where fee accrual alone tips it underwater.*

with a claim that the new 100% is "the safe max." Tracing it, the old comment was factually right:

- Chip → `onAmountChange(((maxBorrowable * 100n) / 100n / WAD) * WAD)` (`StakeManageBorrowCard.tsx:149`)
- `maxBorrowable` = `min(ceiling headroom, simulatedVault.maxSafeBorrowableIntAmount)` (`ManagePositionTakeover.tsx:161-183`)
- `maxSafeBorrowableIntAmount` = `removeDecimalPartOfWad(daiAvailable(collateralValue, debtValue, mat))` (`calculateVaultInfo.ts:59`)
- `daiAvailable` = `collateralValue/mat − debtValue` (`math.ts:116`) — **the liquidation threshold itself, no buffer**

So when collateral binds (the common case — *not* the debt-ceiling case the two updated tests exercise), the 100% chip stages a borrow that lands the vault at ~100% liquidation proximity → `RiskLevel.LIQUIDATION`. The only cushion is the whole-USDS floor: under 1 USDS on a position typically sized in thousands. One stability-fee tick, or any OSM downtick, makes it liquidatable.

Nothing stops submission either. The liquidation-risk guard is scoped to `stakeError` and gated on `skyToFree > 0n` (`ManagePositionTakeover.tsx:139-166`); `borrowError` only checks the debt ceiling and the simulation, and `vat.frob` accepts `tab == ink * spot`, so the simulation passes and Confirm enables.

The defense in the PR body — "the same max the slider's full-right stop already reaches" — is true for the manage flow, but reachable-by-deliberate-drag and reachable-by-one-tap are different risk profiles, and the comps were probably not drawn with this math in mind.

**Suggestion:** either keep a buffer under the chip (stage ~99% of max, or `maxBorrowable − dust`), or extend the `isLiquidationError` guard to borrow mode so the exact-max state disables Confirm with "Liquidation risk too high." If product wants the comp's literal 100% regardless, that's a fine call — but it should be recorded explicitly on the ticket rather than as a comment swap.

### Other issues

**The chip change leaks into the open-position takeover (medium).** `BORROW_PERCENT_CHIPS` is shared with `StakeTakeoverBorrowCard`. The PR scopes itself to the manage comps (1036:213930) and files this under "Manage takeover," but `OpenPositionTakeover.test.tsx` had to be updated too — so the open flow (its own comp set, shipped in #1780) silently changes. Worse there: `shouldShowSlider` requires `debtValue > 0n`, so on a fresh position the slider doesn't exist until an amount is staged — the chip is the only affordance and there's no visible max. Either scope the constant per-surface, or call the open-flow change out explicitly.

**Mode pills re-implement an existing DS primitive (medium).** `StakeManageCard.tsx:47-49` hardcodes:

```
border-[#9ca0e5]/40 bg-gradient-to-b from-[#949aff]/20 to-[#504dff]/20
```

Those are byte-for-byte the existing tokens: `--color-borderBrandDim: rgba(156,160,229,0.4)`, `--color-brand2-start: rgba(148,154,255,0.2)`, `--color-brand2-end: rgba(80,77,255,0.2)` (`globals.css:111-116`). And `tabs.tsx` already ships this exact recipe as `tabsTriggerVariants({ variant: 'pill' })` — same 32px height, same `border-glassBorder` at rest — with a comment stating *"Non-Radix toggle groups reuse these recipes by setting `data-state="active"/"inactive"` themselves."* Reusing it also recovers the DS hover (`hover:border-borderTertiary hover:bg-glassBadge`), the focus ring, and `bg-origin-border` — which the hand-rolled version actually needs, since the translucent active border currently composites over the page rather than the local gradient stop.

**Both borrow and repay lose their max hint (medium).** The PR says the borrow max "stays on the slider's right label," but the slider is conditional on existing debt. A position with collateral and zero debt, opened via the manage takeover, now shows no max anywhere until an amount is typed. Repay is worse: `maxRepayable` is dust- *and* wallet-balance-aware, and that number is now surfaced only reactively through the error ladder. Both comps show "Borrowed:" — worth confirming with design whether dropping the max entirely was the intent or just a relocation.

**Claim chip loses precision on dust (low).** `formatBigInt(amount, { compact: true, maxDecimals: 2 })` — `formatNumber`'s `lessThanSmallest` path now renders anything under 0.005 as `<0.01`, where the previous default gave 4 decimals. The chip only shows when `> 0n`, so small claimables read as `<0.01` rather than `0.0012`. Probably fine, but worth a conscious nod.

**`nextClassName` is now dead (low).** `StakeManageStatCell` still accepts it (`StakeManageCard.tsx:74,81,93`) but no caller passes it — the delta went white→white, which was its whole purpose. Drop it.

**The desktop stat grid is arranged implicitly (low).** `md:grid-cols-[120px_1px_120px_1px_minmax(0,1fr)]` gets its layout purely from how many items each `StatPair` contributes after `md:hidden` dividers drop out of flow. It works out to exactly two rows of `cell│cell│cell` — I checked — but the correctness is invisible: adding one cell, or forgetting an `md:hidden` on a pair divider, silently reshuffles every subsequent cell into the wrong track, with no test to catch it. A comment naming the intended per-row sequence, or explicit `md:col-start-*`, would make the invariant legible.

**Badge recipes are proliferating (low).** This PR adds four near-identical inline pills: the claim chip (`h-5`), `RiskPill` (`h-[18px]`), `RiskBadge` (`h-4`), `UpdatedHourlyBadge` (`h-[18px]`). The DS `Badge` in `components/ui/badge.tsx` is 24px-only, so local recipes are understandable — but an `xs` variant there would be the better home, especially for `UpdatedHourlyBadge`, which is imported cross-file anyway.

**Nit:** `{isStake ? t\`Balance:\` : t\`Staked:\`}{' '}` inside a `flex gap-1` container double-spaces when the skeleton branch renders (explicit space + gap). Drop one.

### Secondary text color — we're using the wrong one

Minor in this PR specifically, but it keeps recurring in our recent PRs: these surfaces should be on the new secondary color, and they're on the legacy one.

It shows up as a visible inconsistency inside a single card, because `StakeTakeoverAmountField` was already moved over in f5544206d ("put the takeover's secondary text on fg-secondary"). So in the manage cards the field label renders `#a1a6c4` while the stat-cell label immediately below it renders `rgba(198,194,255,0.8)` — two different greys, side by side, in dark mode.

Rather than keep fixing this per-PR, we can just swap the value and retire the distinction. Looking at the tokens:

```
--color-textSecondary: rgba(198, 194, 255, 0.8);   /* dark, legacy */
--color-fgSecondary:   #a1a6c4;                     /* dark, DS colors/fg/fg-secondary */

--color-textSecondary: #3b3952;   /* light — comment already says "fg-secondary" */
--color-fgSecondary:   #3b3952;   /* light — identical */
```

Light is **already identical**, so the swap is a dark-mode-only value change: point `--color-textSecondary` at `#a1a6c4`. That fixes all 442 existing `text-textSecondary` call sites at once, versus 147 already on `text-fgSecondary` — far less churn than mass-renaming, and it removes the trap instead of asking everyone to remember which one is current. Probably worth doing as its own small PR (needs a dark-mode visual pass over the app), but it would stop this class of mistake for good.

codewiththeflowxyz:
Thanks for the deep trace — verified every claim against the code and in the browser. Here's where we landed:

**Blocking (100% chip):** Confirmed end to end — `daiAvailable` has no buffer, the whole-USDS floor is the only cushion, and the liquidation guard is indeed scoped to `skyToFree > 0n`, so Confirm stays enabled at the boundary. Agreed the updated tests only cover the ceiling-bound case. We'll fix (buffer under the chip or extend the guard to borrow mode) rather than record a 100%-regardless call.

**Open-flow leak:** Agreed the shared constant changes the open takeover, and `OpenPositionTakeover.test.tsx` moving is the tell. One correction though: the open flow *does* still show a visible max — `StakeTakeoverBorrowCard` kept the `max. X USDS` `topRight` (verified in the browser: it renders above the chips even at zero debt with no slider). It's the manage flow that dropped it. Still needs a scoping decision on the chip constant.

**Max hints (manage):** Confirmed regression — pre-PR the field's `topRight` showed `max. X USDS` unconditionally in both modes; now borrow's max only lives on the debt-gated slider label and repay's is gone. We'll restore or confirm the relocation with design.

**Mode pills:** Agreed — the hexes are byte-identical to the tokens and `tabsTriggerVariants({ variant: 'pill' })` is the right home (also confirmed the `bg-origin-border` difference by measurement). One nuance for the swap: the DS pill measures 34px (`py-2` + `leading-4` + borders), not the hand-rolled `h-8` = 32px the comp draws, so we'll need to reconcile the 2px.

**Claim chip `<0.01`:** Acknowledged as a conscious trade — compact keeps the pill bounded on huge balances; sub-0.005 claimables reading `<0.01` seems acceptable for a chip whose action is "go claim". Happy to revisit if design disagrees.

**`nextClassName`:** Will drop it, though one correction: it wasn't killed by this PR — `git log -S 'nextClassName='` shows no caller ever passed it since the prop landed in `bd822eb86`. Born dead, not made dead.

**Stat grid:** Fair — we'll make the invariant legible (comment naming the per-row sequence, or explicit `md:col-start-*`).

**Badges:** Agreed it's drifting; we'd rather take the `xs` Badge variant as a follow-up than grow this PR.

**Double-space nit:** Checked this one in a live DOM and it doesn't reproduce — the trailing space collapses at the end of the anonymous flex text item, so the skeleton branch renders exactly the 4px `gap-1`. And the space is load-bearing in the value branch, where both strings merge into one flex item and `gap` doesn't apply (removing it renders `Balance:5,000`). Leaving as is.

**Secondary color:** Your swap already landed on `app-redesign` (`252454039`, ~20 min after this review) and is in this branch via the conflict merge — both tokens now compute `#a1a6c4` dark / `#3b3952` light, so the side-by-side grey mismatch is gone with no per-PR churn needed.

codewiththeflowxyz:
**Fixed in f77d6af7a — the 100% borrow chip issue (blocking item from the review).**

It turned out to be a straight regression: APP-426 (`e4a1b6358`, landed July 28) had capped the top borrow chip at 75% for exactly the reason raised here — staging the exact max sets the liquidation price equal to the enforced OSM price, so fee accrual alone tips the position underwater. The restyle then moved the chips back to 25/50/100 to match the comp labels, silently undoing that fix.

Resolution: chips are back to **25/50/75** with the pre-restyle staging math, in both the open takeover and the manage sheet (they share the constant). The pre-restyle tests are restored verbatim, plus a small mock fix (`stabilityFee` as the WAD bigint the real hook returns).

Verified live on a Tenderly fork (10M SKY staked, max 208.33K USDS):

| | Before | After |
|---|---|---|
| Top chip stages | 208,333 (= exact max) | 156,249 (75%) |
| Liquidation price vs protocol price | **$0.025 = $0.025** (born liquidatable) | $0.0187 vs $0.025 |
| Risk pill | High | Medium |

If design wants the comps' 25/50/100 labels back, that needs a deliberate decision about what "100%" should stage — happy to pick that up as a follow-up with them.
fix1-before-after

codewiththeflowxyz:
**Fixed in 9f86aa065 — "Both borrow and repay lose their max hint".**

Confirmed both holes live on a fork: a debt-free position (2M SKY staked, ~41.67K USDS borrowable) showed no max anywhere — the slider whose right label carries it only renders once the position has debt — and in repay, a 50K debt with only 20K USDS in the wallet showed nothing but "Borrowed: 50,000", so tapping 100% staged 20,000 with no visible explanation.

Resolution: the comp's "Borrowed:" line stays, and the max now rides along after it — `Borrowed: 50,000 · max. 20K USDS`. Same semantics as pre-restyle: repay always shows the wallet- and dust-aware `maxRepayable`; borrow shows `maxBorrowable` except below the min-collateral floor, where the warning card owns the state.

Mobile: the Borrowed values render compact below `md` (`Borrowed: 50K → 30K · max. 20K USDS` holds one row at 390px), and the two parts are nowrap-per-chunk, so the narrowest screens break between them — never mid-hint. Desktop keeps full precision.

Covered by four new/extended unit tests (wallet-capped repay max, debt-free borrow with no slider, hint hidden below min collateral, zero-debt repay), verified live on the fork in both themes.

fix2-before-after

codewiththeflowxyz:
**Remaining review items, addressed in five atomic commits.**

**Mode pills → DS recipe (7a5f9df30).** The pills now render `tabsTriggerVariants({ variant: 'pill' })` on plain buttons with `data-state` driven from the active mode — the same non-Radix reuse as `EarnTableFilters`. Verified live: the active border/gradient are the same values via `borderBrandDim`/`brand2` tokens, and the recipe recovers the DS inactive hover (`border-borderTertiary` + `bg-glassBadge`), the `focusRing` focus-visible ring (previously the browser default outline), and `bg-origin-border` compositing. One correction to the review: the recipe is not "same 32px height" — its auto-height box is 34px (16 line + 16 padding + 2 border), which is what every existing pill-tab consumer (wallet drawer, earn filters, portfolio tabs) already renders. Kept the natural 34px rather than overriding back to the hand-rolled 32.

**Dead `nextClassName` dropped (2e9b5d4ff).**

**Stat-grid invariant documented (e3c2c066d).** The comment now names the five-track rhythm (cell│seam│cell│seam│cell), the per-row cell sequence, and why pair 2's divider is `md:hidden` — so adding a cell or dropping a divider is a legible mistake instead of a silent reshuffle. Chose the comment over explicit `md:col-start-*` to keep the markup unchanged.

**Claim chip dust precision (f9cafd89d).** Conscious fix rather than a nod: below 0.01 the chip skips the 2-decimal cap and keeps the small-number 4-decimal default, so a real 0.0012 renders as `0.0012` instead of `<0.01`. Unit test added; large amounts still compact (`123.46M`).

**Balance/Staked double space (df2ebc954).** The label is its own flex item now, so `gap-1` spaces both the skeleton and value branches identically.

**Acknowledged, not done here:** the Badge `xs` variant — agreed it's the right home, but the four local pills currently sit at four heights (16/18/18/20px), so consolidating needs a quick design decision on the canonical XS size; proposing as a follow-up. And the `--color-textSecondary` → `#a1a6c4` repoint — agreed, as its own PR with a dark-mode visual pass, per your suggestion.

iceanddust (approved):

PR #1785, opened 2026-07-31
Ship Circular XX Medium so font weights actually render

iceanddust:
## Why

Font weights have never worked in the app. `font-medium` looked identical to `font-normal`, and `font-semibold` only produced a faint smear. The cause is in `globals.css`:

```css
@font-face { font-family: 'GraphikStd'; src: URL('/fonts/Graphik-Regular-Web.woff2') format('woff2'); }
@font-face { font-family: 'CircleStd';  src: URL('/fonts/CircularXXWeb-Book.woff2') format('woff2'); }
```

Neither face declared a `font-weight` descriptor, so both defaulted to 400 and each family had exactly **one** candidate. CSS font matching resolved every requested weight to that single face. Chrome only synthesises bold at ≥600, so 450/500 were literally pixel-identical to 400, and 600/700 got faux-bold rather than a real cut.

We had shipped only the Book cut since Feb 2025. **`CircularXXWeb-Medium.woff2` was sitting unused in the same licensed Lineto bundle** — the repo's `CircularXXWeb-Book.woff2` is byte-identical (matching sha256) to that bundle's copy, so this is the same licensed package, not a new purchase.

## What the design system actually asks for

Exactly two weights, from the `5. Typography` collection:

| Variable | Value | Family |
|---|---|---|
| `font-weight/body` | 400 | Graphik (Body styles) |
| `font-weight/label` | 500 | Circular (all Label + Heading styles) |

So Graphik needed nothing — its single 400 face was always correct. Circular Medium was the only missing resource.

## Changes

**1. Ship Medium + declare real weights** — `CircularXXWeb-Medium.woff2` added; all three faces get an explicit `font-weight` and `font-display: swap`.

Circular XX Book is **450** internally, not 400 — declaring that makes the existing `--font-weight-custom-450` token resolve for the first time. `font-normal` (400) still lands on Book per CSS weight matching.

**2. Portfolio footer stats onto the comp's styles** — the values used `text-lg font-medium`, which missed Figma `1030:58701` three ways: Graphik instead of Circular (no `font-circle`), 18px instead of 16 (18 is Label 4's *line height*), and no `-0.32px` tracking. Labels also drop 14px → the Body 6 12/18 the comp specifies.

Worth noting: the comp's label/value contrast comes from a **family switch** (Graphik Regular → Circular Medium), not a weight change. That's why no amount of `font-weight` reproduced it.

## Verification

In-browser on the design-system page, dark mode:
- All three faces report `loaded`: `GraphikStd w=400`, `CircleStd w=450`, `CircleStd w=500`
- Same string measures **656.14px at 450** vs **669.95px at 500** — two genuinely different faces, not synthesis
- The emitted utilities resolve correctly: `font-circle font-medium` → Medium, `font-circle font-custom-450` → Book, bare `font-medium` → Graphik 400

Gates: typecheck clean, prettier clean, 2210/2211 unit tests pass. The one failure (`EarnFeaturedCards` Pendle maturity) and the 3 lint errors (`ProductCard.test.tsx`) both reproduce on a clean `app-redesign` — pre-existing, untouched here.

The portfolio card itself wasn't visually QA'd in-app: it's wallet-gated and the vnet in `.env` is expired. The CSS behaviour and the exact class recipe were verified directly instead.

## Follow-ups (deliberately not in this PR)

- **~121 usages ask for weights the design system doesn't define**: 101 `font-medium` on Graphik and 20 `font-semibold` (there is no 600 anywhere in the system). Each is either a missing `font-circle` or a `font-medium` that should be dropped. Shipping Medium doesn't touch these.
- `label4`/`label5`/`label6` are module-private in `ui/table-cells.tsx`. Promoting them to shared typography primitives would stop surfaces hand-rolling their own recipe — which is exactly how this card drifted.

## For design / legal

- **Licence scope is `sky.money` + subdomains, <5M monthly pageviews.** Vercel previews serve these self-hosted fonts from `*.vercel.app`, outside that grant. Worth someone confirming rather than assuming.
- **The DS names the heading family `Circular Std`; we're licensed for Circular XX.** Different Lineto products. XX is what we can legally ship, so the variable should be relabelled — and if the comps were drawn in Std, expect minor letterform differences vs production.
- **The body family in Figma is `Graphik Trial`** — the trial cut, not licensed for production. We ship real Graphik, so this is likely just Figma-file sloppiness, but the comps may not be metrically identical to what ships.
- **The licensee is "Maker Growth"** — worth confirming that entity still covers us post-rebrand.


iceanddust:
## Added: `antialiased` on `body` (3rd commit)

Shipping the real Medium 500 cut surfaced a second issue — on **macOS** Chrome, headings and labels rendered a visible notch heavier than the comps. Chrome applies a contrast/gamma boost that Figma's canvas doesn't.

Diagnosed rather than assumed, because two different causes produce the same symptom: (a) macOS rendering, or (b) a Circular XX vs Circular Std family mismatch (the DS names `Circular Std`; we're licensed for XX). A screenshot can't tell those apart. Three findings separate them:

1. **Lineto's current "Circular" *is* the XX one** — so the DS's `Circular Std` is a stale label, not a different product. No family mismatch.
2. **This branch already renders correctly on Windows**, with identical font files and no override — and `-webkit-font-smoothing` is a no-op outside macOS. So the font data is right; the deviation is macOS rendering only.
3. **Lineto ship their own Circular specimens with `antialiased` enabled.**

So the fix only changes the platform that's wrong and is inert on the one that's already right.

**Caveats:** thinning is more pronounced on non-retina displays (worth a look if anyone runs a 1x external monitor), and lighter strokes read as slightly lower contrast — no change to computed WCAG ratios, but relevant for small secondary text.

This is a separate commit (`e0a169aa8`) and can be dropped independently of the font fix.

**Housekeeping for design:** since XX is the current Circular, the `Font Family/Headings` variable should be relabelled from `Circular Std` to `Circular XX` so this doesn't get re-litigated. The `Graphik Trial` and `Maker Growth` licence questions in the description are unaffected and still open.

Gates after this commit: typecheck clean, prettier clean, 2210/2211 unit tests (same pre-existing `EarnFeaturedCards` Pendle-maturity failure as `app-redesign`).

inclinedmind (approved):

PR #1786, opened 2026-07-31
Resolve weight requests the design system doesn't define

iceanddust:
**Stacked on #1785** (`redesign/circular-medium-weight`) — review that one first; this PR's diff is only the two commits on top.

## Why

#1785 shipped Circular Medium 500 and made `font-medium` real. This is the follow-up it named: the ~121 sites asking for a weight the design system doesn't define.

The DS has exactly two weights — `font-weight/body` 400 on Graphik, `font-weight/label` 500 on Circular. These sites requested 500 or 600 on **Graphik**, which has neither cut. They silently rendered at 400, or (for `font-semibold`) as Chrome's synthetic bold. Shipping Medium didn't touch any of them, because none of them said `font-circle`.

Each site is resolved to whichever the comp actually wants:

- **Label / value / badge / title / tooltip heading / amount input → `font-circle`**, so it lands on the real Medium cut. Covers stake (largest group), portfolio, convert, savings, claim, balances, toasts, chart tooltips, and the shared widget primitives.
- **Inline emphasis inside Graphik prose → drop the weight.** The `ProductDetailTemplate` "Learn more" link and the `AllocateStablecoinsBanner` rate span already carry emphasis in colour, and there is no Graphik 500 to promote them to. Removing the class is a **visual no-op today** — it was never rendering — and stops the code claiming a weight that doesn't exist.

**Family and weight only.** No sizes, tracking or line-heights changed; that would have ballooned the diff into a re-spec.

## One real bug found

`components/ui/button.tsx` — the DS size recipes (`xl/l/m/s/mini`) each add `font-circle`, but `defaultVariants` sets `size: 'default'`, which doesn't. So any `<Button>` with no explicit `size` prop rendered in Graphik. `font-circle` now lives on the base.

## Deliberately untouched

| Site | Reason |
|---|---|
| `CardWidgetTitle`, `AccordionWidgetTrigger`, `TableFooter`, `SelectLabel` | Exported, **zero consumers**. Dead. |
| `buttonWidgetVariants` | One consumer (`pagination.tsx`), itself legacy. |
| `widgets/components/ui/input.tsx` | Its `font-medium` is `file:font-medium` — shadcn file-input boilerplate, not the field's text. Rewriting it would have produced broken output. |
| `widgets/TradeWidget`, `modules/analytics`, `pages/SealEngine` | Out of the agreed scope: no route mounts Trade/Upgrade, and the others are uncomped internal pages. |

The four dead exports are a good candidate for a separate deletion PR — they're carrying stale styling nothing renders.

## Verification

- typecheck clean, prettier clean, lint at the pre-existing baseline (3 errors in `ProductCard.test.tsx`, untouched here)
- **2210/2211** unit tests — the one failure is the `EarnFeaturedCards` Pendle date-bomb, which reproduces on a clean `app-redesign`
- Checked the bulk transform for damage: no variant-prefixed classes (`hover:font-circle` etc.) and no doubled classes in the final diff. One `font-circle font-circle` did appear mid-pass on a `font-semibold` line and was fixed before commit.
- JSDoc comments were guarded so prose mentioning `font-medium` wasn't rewritten.

Not browser-verified: most of these surfaces are wallet-gated and the vnet in `.env` is expired. The changes are family swaps on classes whose weight was previously inert, so the risk is a surface reading heavier than intended rather than breaking — worth a pass on the Vercel preview, particularly stake, which owns the largest share of the diff.

## Still open from #1785

The `Font Family/Headings` variable should be relabelled `Circular Std` → `Circular XX`, and the `Graphik Trial` / "Maker Growth" licence questions are unaffected by this PR.

### DIFF
--- a/apps/webapp/src/components/product/ProductDetailTemplate.tsx
+++ b/apps/webapp/src/components/product/ProductDetailTemplate.tsx
@@ -97,16 +97,20 @@ function ProductTitleIcon({ token }: { token: ProductDetailToken }) {
   );
 }
 
+/* font-medium is load-bearing: font-circle alone falls to `normal` (400), which
+   resolves to Circular Book 450 — the DS has no Circular Book, every Headings-family
+   style is weight 500 (font-weight/label). */
 function SectionHeading({ className, children }: { className?: string; children: ReactNode }) {
-  return <h2 className={cn('text-text font-circle text-lg', className)}>{children}</h2>;
+  return <h2 className={cn('text-text font-circle text-lg font-medium', className)}>{children}</h2>;
 }
 
 /* M6.3 section-heading scale (486:20706): Details/About step down to Label 4,
-   Transactions steps up to Heading 6; both return to the shared 18px at md. */
+   Transactions steps up to Heading 6; both return to Label 3 (18/22, -0.36px,
+   comp 859:35722) at md. */
 const minorHeadingClasses =
-  'text-base leading-[18px] tracking-[-0.32px] md:text-lg md:leading-normal md:tracking-normal';
+  'text-base leading-[18px] tracking-[-0.32px] md:text-lg md:leading-[22px] md:tracking-[-0.36px]';
 const majorHeadingClasses =
-  'text-xl leading-[22px] tracking-[-0.4px] md:text-lg md:leading-normal md:tracking-normal';
+  'text-xl leading-[22px] tracking-[-0.4px] md:text-lg md:leading-[22px] md:tracking-[-0.36px]';
 
 function DetailsSection({ title, details }: { title?: ReactNode; details: ProductDetailRow[] }) {
   return (
--- a/apps/webapp/src/components/product/ProductDetailTemplate.tsx
+++ b/apps/webapp/src/components/product/ProductDetailTemplate.tsx
@@ -143,12 +143,7 @@ function AboutSection({ title, about }: { title?: ReactNode; about: ProductDetai
       <div className="text-fgSecondary font-graphik text-sm leading-[22px]">
         {about.body}
         {about.learnMoreHref && (
-          <a
-            href={about.learnMoreHref}
-            target="_blank"
-            rel="noreferrer"
-            className="text-fgBrand ml-1 font-medium"
-          >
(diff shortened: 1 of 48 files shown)


iceanddust:
## Added: product Details values on Label 4 at desktop (3rd commit, `990836ac9`)

Desktop comp [859:35723](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=859-35723&m=dev) specs every Details row as a Body 5 label (Graphik 14/22) against a **Label 4 value (Circular Medium 16/18, −0.32px)**. The value span was doing:

```
text-text font-circle md:font-graphik ... md:text-base md:leading-normal md:tracking-normal
```

Circular on mobile, then **switching to Graphik at `md`** — exactly where the comp calls for Circular Medium — with `normal` leading and tracking instead of 18px / −0.32px. From `md` up, every data point on every product page rendered in the body face at the wrong metrics.

**Not a design decision.** `git log -S` traces the `md:` overrides to the M6.3 mobile build (`b0e03b569`, APP-378), which added the mobile Label 5 treatment and pinned the *pre-existing* desktop styling behind `md:` rather than checking it against the desktop comp. Mobile was comp-verified; desktop was inherited.

It was also effectively invisible until #1785 — with only the Book cut shipped, Circular and Graphik at `font-medium` were close enough that a family switch didn't read as wrong. Shipping the real Medium is what exposed it.

**Scope:** fixed in the shared `ProductDetailTemplate`, so all six product pages inherit it (savings, vaults, fixed/Pendle, rewards, stUSDS, stake). Checked all 34 consumers — none styles its own row values.

**Mobile untouched:** `text-sm leading-4 tracking-[-0.28px]` is the M6.3 Label 5 step-down against its own comp (486:20706), pixel-verified when built.

Labels also take the comp's 22px line height at `md` (`leading-normal` on 14px gives 21px). Same rows, same comp — easy to drop separately if you'd rather keep this to the values.

Gates after this commit: typecheck clean, prettier clean, **2210/2211** (same pre-existing Pendle date-bomb).

inclinedmind (approved):

PR #1788, opened 2026-07-31
Add page transitions on route navigation

iceanddust:
Prototype of the page-transition motion comp ([Figma 1598:75512, view=motion](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1598-75512&focus-id=1598-75512&view=motion&m=dev)), for evaluation on the Vercel preview.

## What the comp specifies

Read off the timeline keyframes (6s loop cycling 4 pages, one `x` + `opacity` track each):

| Layer | Property | From → To | Duration | Curve |
|---|---|---|---|---|
| Outgoing page | `opacity` + `x` | 1→0, `0 → -120px` | 300ms | `cubic-bezier(.77,0,.175,1)` |
| Incoming page | `opacity` + `x` | 0→1, `+120px → 0` | 300ms | same |
| Navbar active pill | `opacity` crossfade | 1→0 / 0→1 | ~100ms | same |
| (Page 3 only) chart | `width` 0→760 | — | ~600ms | same, ~150ms into the page enter |

Two things the numbers settle:

1. **Direction is constant** — always out to the left, in from the right, regardless of which nav item is clicked. No back-goes-the-other-way.
2. **The pages overlap.** The incoming page starts ~100ms into the outgoing one, so ~200ms has both on screen. It is explicitly a crossfade, not a sequential out-then-in.

## Approach

View Transitions API, via the router's `defaultViewTransition`, rather than `AnimatePresence` around the `<Outlet/>`.

The deciding factor is point 2: the outgoing page has to stay painted for 300ms after React unmounted it. Wrapping the `Outlet` in `AnimatePresence` has a well-known failure here — the exiting copy still contains a live `<Outlet/>`, so when router context updates it re-renders **the new page** and you watch the new page fade out. Working around that means snapshotting the matched component by hand, and you still have to keep two full pages in document flow without collapsing scroll height, now that pages scroll on the document (G5). Letting the browser snapshot sidesteps all of it.

We implemented and removed View Transitions during the route-splitting refactor, but that was because the exit fade was judged unnecessary once instant nav felt fine — not because it didn't work. The one real blocker (the root snapshot swallowing clicks for the whole animation) was solved then, and that fix is carried over here: `:root { view-transition-name: none }` so only the page group participates.

Three deliberate deviations from the comp:

- **Travel is capped** at `min(120px, 8vw)`. The comp's 120px is measured on a 1512px frame; at 390px wide a literal 120px reads as the page being flung off screen.
- **Reduced motion** (`prefers-reduced-motion: reduce`) keeps a 150ms crossfade but drops the translate. Per WCAG 2.3.3 the vestibular trigger is large-area movement, not opacity — removing the transition entirely also loses the "the page changed" signal and reads as a flash.
- **The navbar pill is not animated.** The comp fakes its state change by crossfading two whole button instances; in CSS the active state is a gradient background, and gradients don't interpolate. Mobile already has a nicer `layoutId` spring pill. Left alone — happy to do it properly as a follow-up if design wants it.

The chart wipe is page content rather than page transition, so it isn't in scope here either.

## Which navigations animate

`types` is a callback so it can veto: navigations that only rewrite search params — network switch, stake urn index, balance filters — return `false`, which skips `document.startViewTransition` entirely and leaves them as instant swaps. Only real path changes get tagged `page`.

All the animation CSS is scoped to `:active-view-transition-type(page)`, so a browser without view-transition-types support matches nothing and degrades to today's instant swap.

## Verification

- `pnpm typecheck` clean.
- `pnpm lint` — 3 errors, all pre-existing in `ProductCard.test.tsx` (identical count on base).
- `pnpm --filter webapp test` — 2210/2211. The one failure is `EarnFeaturedCards.test.tsx`, the known date-bomb (a Pendle maturity date that is now in the past); fails identically on base.
- Verified live in the browser: `startViewTransition` fires with `types: ['page']`, exactly one element carries `view-transition-name: page`, the root name is suppressed, and all 14 CSS rules parse and survive (no selector silently dropped).

**Not verified: how it actually looks.** Chrome aborts view transitions in a backgrounded tab (`visibilityState: 'hidden'`), and the automation tab could not be brought to the foreground, so no frame of the animation was ever observed locally. That is the reason for the preview.

Two specific things worth watching on the preview, both consequences of naming a tall, document-scrolled element:

1. **Height morph.** Old and new pages differ in height, and the group's box interpolates between them, which can stretch the outgoing snapshot on a tall → short navigation (`/earn` → `/convert` is the sharpest case). `object-fit: none` + `object-position: top left` is in there to prevent it, unconfirmed.
2. **Scroll offset.** With `scrollRestoration` resetting to top, navigating from a scrolled position means the old snapshot's viewport-relative origin differs from the new one, which could show up as an unwanted vertical slide.

If either looks wrong, the fallback is to snapshot the root and lift the header, background and mobile navbar out with their own names — the root snapshot is always exactly viewport-sized, so neither issue can arise. That trades in the click-swallowing problem, which is why it isn't the first choice.


iceanddust:
## Fixed: the flash on stake / convert / product pages

Reproduced and traced. The transition CSS was never the problem — **two view transitions were starting for a single navigation**, and the browser skips the earlier one when a new one begins. A skipped transition still applies its DOM update immediately, which is the flash; the second transition then animates normally.

Measured with Playwright (headless Chromium reports `visibilityState: 'visible'`, so transitions actually run there), instrumenting `document.startViewTransition` and sampling per animation frame:

```
portfolio -> /earn   (good)  1 call   ready OK
portfolio -> /stake  (bad)   2 calls  #1 ready REJECTED AbortError: Transition was skipped
                                      #2 ready OK
portfolio -> /convert (bad)  2 calls  same
```

Both calls came from an identical stack — `loadMatches`' `onReady` — meaning `load()` ran twice for one navigation. There was only one `pushState`, so it wasn't a second history entry.

### Root cause

`useAppSearchParams`' setter called `router.navigate` unconditionally. Navigating to the search string already in the URL makes TanStack skip the *history write* but still run a full `load()` — and that second load fires its own view transition.

`useAppOrchestration` re-validates search params whenever the intent changes:

```js
useEffect(() => {
  setSearchParams(params => validateSearchParams(params, intent), { replace: true });
}, [searchParams, intent, newChainId]);
```

It arrives with nothing to change, so this fired on every route carrying its own intent — stake, convert, the product pages — but never between two routes sharing one. `pathToIntent` maps both `/portfolio` and `/earn` to `BALANCES_INTENT`, so that pair never re-ran the effect, which is exactly why it was the one path that looked flawless.

Confirmed by disabling the effect and rebuilding: the second call disappeared on all routes.

### Fix

`setSearchParams` now bails when the result equals the current search string (compared key-sorted, so a bare reordering doesn't count). This also drops one redundant `load()` per module navigation, which predates the transitions.

After the fix, every route fires exactly one transition and the new content only ever appears while the transition is active:

```
portfolio -> /stake     1 call, ready OK
    0–  48  vt=off     /portfolio   (old page)
   76–  76  vt=off     /stake       (path updated, old content still painted)
   93–  93  vt=ACTIVE  /stake       (transition starts)
  192– 636  vt=ACTIVE  /stake       (new content, animating)
  637–1651  vt=off     /stake       (done)
```

Previously the new content appeared at the very first `vt=ACTIVE` frame — the skipped transition's instant DOM update.

Verified on a local production build for `portfolio -> earn`, `portfolio -> stake`, `portfolio -> convert`, `stake -> convert` and `earn -> portfolio`. Product-page routes couldn't be exercised locally — without the geo API reachable, `requireModuleEnabled` falls back to the restrictive config and bounces them to `/portfolio` — but they run the same intent-change code path as stake and convert.

Gates unchanged: typecheck clean, 2210/2211 (the one failure is the known `EarnFeaturedCards` date-bomb, fails on base), lint at the same 3 pre-existing errors.

### Still worth a look on the preview

The two geometry risks from the original description are unaffected by this and remain unverified: the group's box morphing between differing page heights, and navigating from a scrolled position.

iceanddust:
## Both geometry risks fixed — the page now rides the root snapshot

`f6daafd4c`. Switched from naming the page wrapper to capturing the page as the **root snapshot**.

### Why the vertical slide happened

Snapshots are positioned in viewport coordinates, and the transition group was the page wrapper — a tall in-flow element. Scrolled down 1200px, its top sits at −1200px relative to the viewport; after `scrollRestoration` jumps to the top, the new one's top is at 0. The browser auto-generates a group animation interpolating that box, so it translated by the whole scroll delta alongside the intended horizontal slide.

The root snapshot is always exactly the viewport and contains what was literally on screen, so the scroll offset is baked into the image and there is nothing left to interpolate. This kills risk #1 as well — both root snapshots are viewport-sized, so there is no box to morph between pages of differing height.

The chrome that must hold still (header, `.app-background`, mobile navbar) now carries its own `view-transition-name`, which lifts it out of the root snapshot into its own group.

### Three things that had to be handled

**Group stacking is not inherited.** Groups are siblings painted in capture order; the document's z-index does not carry across. `.app-background` (z-index −10 in the page) painted *over* the page snapshot and blanked the content area for the entire animation. The intended order is now restated explicitly on the groups.

**The document had no background colour** — `html` and `body` both computed `rgba(0,0,0,0)`; `.app-background` was the only painter, and while it is lifted into the pseudo tree the document has none. Every region the snapshots didn't cover fell through to the browser's default white. `html` now carries `--color-pageBackground` as a backstop, which also covers overscroll and the gap before first paint.

**Click-swallowing is fixable after all.** Snapshots hit-test in the top layer, so a click during the ~400ms animation landed on a pseudo-element and died on `<html>`. `pointer-events: none` is now set on *every* pseudo — `::view-transition`, `-group()`, `-image-pair()`, `-old()`, `-new()` — not just old/new. The incomplete form is what made this look unfixable when view transitions were first tried here.

### Measured

Playwright against a local production build, dark theme, `prefers-reduced-motion: no-preference`:

```
TEST 1  vertical drift (scrolled 1200px -> new route)
  root-group translateY: min 0 max 0            -> none
  groups: root z=1, shell-background z=0, shell-header z=2

TEST 2  click fired 150ms into the animation
  transition still running when 2nd click fired: true
  landed on: /stake                              -> registered

TEST 3  filmstrip, animation slowed to 4s
  dark background, header sharp and static, both page snapshots crossfading
```

And one `startViewTransition` per navigation on every route, so the flash fix from the previous commit is intact.

Gates unchanged: typecheck clean, 2210/2211 (the known `EarnFeaturedCards` date-bomb), lint at the same 3 pre-existing errors.

### Note

`shell-navbar` reports as not captured at desktop widths, which is correct — it's `desktop:hidden`. The mobile bar hasn't been exercised at a phone viewport yet, so that's worth a look on the preview alongside the general feel.

iceanddust:
## Boot animation + background fixed; merged with `app-redesign`

`9f3796628`, on top of a merge of `origin/app-redesign` (`a6aa71d17` — one trivial conflict in `globals.css`, both sides added a rule above `body`, kept both).

### Animation on page load

The router derives `pathChanged` as `fromLocation?.pathname !== toLocation.pathname`, and `fromLocation` is **undefined on the first load** — so `undefined !== '/earn'` reported true and the app animated itself in on boot. The `types` callback now requires a previous location:

```ts
types: ({ pathChanged, fromLocation }) => (fromLocation && pathChanged ? ['page'] : false)
```

Measured: zero `startViewTransition` calls during initial load, still exactly one per real navigation.

### The background

I had this wrong, and the fix is the opposite of what I did last time.

**The root snapshot always includes the document canvas, so it is opaque.** Anything given its own `view-transition-name` can therefore only ever sit *below* it and is never visible. Naming `.app-background` hid the image layer for the entire animation and left a flat colour sliding in its place — which is what read as the background transitioning, and why it looked odd.

Diagnostic that settled it: painting the group red and forcing `z-index: 99` showed the real background image (so the group captures fine); at `z-index: 0` it was invisible — occluded by the root snapshot.

So `.app-background` goes back to being unnamed and rides inside the root snapshot, travelling with the page. The page colour is uniform, so its movement is invisible; only the soft image layer shifts. `html` keeps `--color-pageBackground` so the strip the outgoing page uncovers isn't browser-default white — `.app-background` is `position: fixed` and doesn't cover that strip. It also fixes overscroll and the gap before first paint.

Only the header and the mobile navbar are named now.

**Trade-off worth your call:** the background image now moves with the page rather than staying perfectly still. Keeping it still is not achievable with a root snapshot, for the reason above — the alternative would be going back to naming the page element and compensating for the scroll delta in JS, which reintroduces the complexity the root snapshot removed. My read is that the whole page including its background sliding is coherent, and the image is soft enough that the shift is subtle, but you're the judge on the preview.

### Verified on a production build

```
initial load          0 transitions
per navigation        1 transition (earn / stake / convert)
vertical drift        root-group translateY min 0 max 0    -> none
click mid-animation   landed on /stake                     -> registered
mid-transition frame  real nebula background, header static, page crossfading
```

Gates: typecheck clean, 2214/2215 (the known `EarnFeaturedCards` date-bomb), lint at the same 3 pre-existing errors.

iceanddust:
## Root-snapshot approach reverted — it was the cause of both background problems

`326fd5af7`.

Both issues trace to the same decision. **The root snapshot always includes the document canvas**, which has two consequences I ran into one at a time:

- Whatever is in that snapshot travels with the page → the background moved with the navigation (dark mode).
- Nothing lifted out of it can ever show, because it's opaque → the `html` background-colour I added to stop the uncovered strip going white ended up painting over the gradient layer → flat light mode.

So the fix isn't another patch on top; it's going back to capturing **only the page wrapper**. An uncaptured element keeps painting live in the DOM for the whole animation, so the background, header and navbar simply cannot move or change appearance. No `html` background is needed either — `.app-background` is still there painting the entire time, which is why light mode renders correctly again.

That structure had exactly one flaw, the vertical slide, and it's now fixed directly instead of by changing what gets captured: the group's auto-generated animation interpolated the wrapper's viewport-relative box, and `scrollRestoration` jumps the document to the top on a new route, so it dragged the whole scroll delta. The group is pinned, and the snapshots are placed explicitly — the router writes the pre-navigation scroll position to `--vt-scroll` in the same synchronous step that starts the transition, and the outgoing snapshot (which is the entire tall element) is offset by it so it shows the part that was actually on screen.

### Measured on a production build

```
background static     .app-background rect during transition: ["0,0"]  -> never moves
vertical drift        group transform every frame: matrix(1,0,0,1,80,76) -> constant, no interpolation
light mode            gradient + diagonal streak render again (matches the reference)
initial load          0 transitions
rapid navigation      4 nav clicks at 30ms / 60ms / 80ms gaps
                      history: ["/earn","/stake","/convert","/portfolio"] -> all four register
```

On the nav blocking: I could not reproduce a dropped click at any gap I tried, down to 30ms, with the complete `pointer-events` rule on every pseudo. If it still happens for you, it would help to know whether the click does nothing at all or the page changes but the active pill lags — those are different causes and I'd chase the second one differently.

Gates: typecheck clean, 2214/2215 (the known `EarnFeaturedCards` date-bomb), lint at the same 3 pre-existing errors.

iceanddust:
Fixed a regression this PR introduced, reported in review: the stake takeovers rendered with no scrim — the live page read straight through them.

`.page-transition`'s `view-transition-name` **breaks `backdrop-filter` for descendants**, and `TakeoverShell` was the one modal-class overlay rendered inside the page rather than portalled, so its `bg-modalOverlay` + `backdrop-blur-[100px]` stopped painting entirely.

Confirmed by toggling `view-transition-name` off on the wrapper at runtime with the takeover open — the scrim came back immediately. (An earlier hypothesis, that the name creates a containing block for fixed descendants, was wrong: the shell measured exactly viewport-sized.)

Fixed in 141e661e1 by portalling the shell to the body, which the dialog and sheet overlays it shares its scrim recipe with already do. Worth watching for any other `fixed` + `backdrop-filter` element rendered inside the page wrapper.

inclinedmind:
Reviewed — the core mechanism looks solid (root-name suppression, hit-testing, scroll compensation, reduced motion, and the no-op search-write race are all genuinely handled; verified against the installed `@tanstack/router-core` internals). Two findings worth addressing, two nits:

**1. The "degrades to instant swap" claim doesn't hold for VT-capable browsers without transition-*type* support** (Chrome/Edge 111–124, Safari 18.0–18.1, Firefox 144–146). TanStack detects types with `CSS.supports('selector(:active-view-transition-type(a))')`, and when that's false but `document.startViewTransition` exists it still starts a transition in callback form. In those browsers:

- the `types` callback never runs, so its vetoes don't apply — search-param-only navigations *and the initial boot render* get a transition
- `--vt-scroll` is never written, so navigating from a scrolled position crossfades the page top instead of the visible region
- the result is the UA default ~250ms crossfade, not an instant swap

Cheap hardening that mirrors TanStack's own detection exactly: wrap the `.page-transition { view-transition-name: page }` rule in `@supports selector(:active-view-transition-type(page))`. With no named element and the root suppressed, those browsers capture nothing and genuinely degrade to an instant swap.

**2. Back/forward navigation can break the "incoming page is at scroll 0" assumption.** `scrollRestoration: true` restores the previous offset on history navigations, and the `types` callback animates path-changing back/forward navs like any other. When the restored offset is non-zero, the old snapshot's `--vt-scroll` compensation is off by that amount, so the outgoing page appears shifted during the transition (or the scroll jump fights the transition's end). Easy repro to try on the preview: scroll down on `/earn`, navigate away, press browser back. If it looks bad: exclude history navigations from the `page` type, or write a second variable for the restored offset.

**Nits:**
- `globals.css` (~line 417): the pre-existing comment "`antialiased` = -webkit-font-smoothing" accidentally lost the space after `=`.
- `::view-transition-group(page)` uses `height: 100vh`; `100dvh` would pin more precisely under mobile dynamic toolbars (harmless as-is since the pseudos are `pointer-events: none`).

No performance concerns — the `types` callback runs once per navigation, and the `setSearchParams` bail actually removes a redundant `load()` that used to fire on every module-route intent change.

*Written by Claude, posted on behalf of inclinedmind.*

inclinedmind:
Also, should it slide to the right side when switching from convert to stake for example, or is it supposed to always slide to the left?

PR #1814, opened 2026-08-05
APP-456: Jacek QA 04.08.2026

iceanddust:
Closes [APP-456](https://linear.app/ekliptyka/issue/APP-456/jacek-qa-04082026).

Eight items; two commits.

### 1 — Footer
`PageFooter` on every page, reproducing the landing page's footer: the legal disclaimer over the bottom row. Two typographic tiers, per the spec on the issue — disclaimer at 10px on `fg-secondary`, copyright and links at 12px on `fg-primary`.

Copy is the landing footer's, verbatim, cut to its first two paragraphs (Kacper: "just first 2 paragraphs will be enough"); its third, FORWARD-LOOKING STATEMENTS, is left off. The links come from `VITE_FOOTER_LINKS`, filtered to the legal documents — Bug Bounty is in that var but isn't legal, so it stays in the More menu and off the footer.

The dead `FooterLinks.tsx` this replaces was removed — nothing rendered it.

### 2 — Header padding
The desktop bar was 8px above and below the pill row. Figma ([Navbar 1881:51590](https://www.figma.com/design/1aCQfCwuGx90hVwGcD2ZLS/Sky-App--UI?node-id=1881-51590&m=dev)) is an 88px bar around a 40px row, i.e. 24px. Browser-measured at 1512px: bar 88px, pills at y=24.

### 3 — Bundled-transaction legal notice
Kacper picked option C. The `/batch-transactions-legal-notice` route and page are gone; `BATCH_TX_LEGAL_NOTICE_URL` now points at the docs.

**Answering the question in the issue:** docs links are hardcoded `https://docs.sky.money/...` strings at each call site — there is no env var for the docs host (`VITE_FOOTER_LINKS` covers the legal links only, and `ALLOWED_EXTERNAL_DOMAINS` allow-lists the host). So this is a plain constant.

⚠️ **The dedicated docs page does not exist yet.** The constant currently lands on `https://docs.sky.money/legal-terms` — the document the notice itself cited — with a TODO naming the swap. This is the "let's make sure we won't forget to add it" note from the thread.

### 4 — Careers
Careers is the deployment's only `highlight` entry in `VITE_FOOTER_LINKS`, and that flag only ever existed to promote it. `getFooterLinks` now drops highlighted entries, so the link disappears everywhere without waiting on a Vercel edit.

⚠️ Worth deleting the entry from the Vercel `VITE_FOOTER_LINKS` var too, so the config matches the code.

### 5 — Charts
Audited every chart in the app. Three findings:

**5.2 — BA Labs paging (the Total USDS and DAI chart).** `/overall/historic/` caps a response at **1000 rows** whatever `p_size` asks for, and advertises the remaining 1458 through `count` and a `next` link built with an **`http://`** scheme the browser blocks as mixed content. So `?p_size=9999` silently returned the most recent 1000 days and the All-time chart started Nov 2023.

`fetchBaLabsPages` walks its own https page URLs, bounded by the caller's `p_size` budget — a hook asking for 1 row still makes exactly one request; the All-time chart makes three. Applied to all six `p_size` callers. Browser-verified: All now starts **Nov 2019**.

**5.1 — Staking reward rate.** The series plotted only the farm with the highest current rate. Farms are added over time, so that truncated All at the SKY farm's launch (2025-11-03) — later than 1Y, and later than the borrow-rate and TVL series on the same card, which read one engine-wide endpoint. It now plots the best rate available on each day across all farms, which spans back to 2025-05-19 and by construction still agrees with the hero on the newest day.

**Also found: the stUSDS detail chart.** It asked `/overall/historic/` with no `p_size` at all, so it only ever got the endpoint's 100-row default — its 1Y and All ranges were clipped to ~100 days. It now takes its row budget from the range picker like every other detail chart.

**Also found: the Total USDS and DAI chart's 1Y and All had seven points.** `useParseTokenChartData` — used only by that chart, on the portfolio page and in the marketplace — flattened those two ranges to seven equidistant samples, drawing a six-segment polyline where every other chart in the app (all on `useParseTvlChartData`) plotted the real daily shape. It now samples on the same timeframe interval as the rest: daily, weekly on 1Y.

### 6 — Header blur
Not fixed by the taller bar: the band got taller but content crossing it stayed legible, because the old mask feathered out across most of the bar's height. A mask fades the blurred layer's **opacity**, not its radius, so over a tall band the sharp original shows through underneath and text reads as a double image rather than as something going out of focus.

The fix keeps one layer and confines the fade to the bottom quarter (the recipe @inclinedmind prototyped in Chrome):

```css
backdrop-filter: blur(20px);
mask-image: linear-gradient(to top, transparent 0%, #000 25%);
```

Content is sharp where it meets the bar's bottom edge, softens as it rises through that 22px, and is fully diffuse behind the logo and pills. The partial-opacity band still overlays a blurred copy on the sharp original, but over 22px right at the boundary that reads as the fade it is meant to be.

It lives on `HeaderBlur`, a real element rather than a pseudo-element, so it shows up in the inspector under its own name — an intermediate version of this had a hard layer edge just above the nav buttons that was reasonably mistaken for a header padding bug. The header box itself measures exactly to the comp throughout: 88px, 24px above and below the 40px pill row.

Verified both themes at 1512px, watching a heading travel up through the bar.

### 7 — Morpho vault withdrawal
Root cause: **Morpho V2 vaults answer `maxDeposit`, `maxWithdraw` and `maxRedeem` with `0n` for every account**, even while deposits and withdrawals are wide open. Verified on-chain against three real holders of the USDS Flagship vault — all three hold shares, all three read `maxWithdraw = 0`.

The widget's supply/withdraw pane already worked around this for Morpho; the product-page modal (`useVaultTransactionForm`) read the contract values directly, so it showed a supplied position with `Balance: 0` and rejected every amount. `shares: maxRedeem` also meant a Max withdrawal would have redeemed nothing.

Per "should be something reused between all of the vaults", the provider split now lives in `computeVaultLimits` and both surfaces call it, so they can't drift again.

### 8 — External link notice
The "Leaving Our Website" interstitial is gone, along with all its support: the modal, the `externalLinkModal*` state and `onExternalLinkClicked` on `ConfigContext`, the `skipConfirm` prop, and the `onExternalLinkClicked` prop threaded through ~40 widget files. `ALLOWED_EXTERNAL_DOMAINS` stays — `sanitizeUrl` still uses it to allow-list env-provided URLs.

---

### Verification

`pnpm typecheck` clean, `pnpm lint` 0 errors, fast suite 2268 passed. The one red test — `EarnFeaturedCards` asserting a Pendle maturity of `18 Jun 2026` — fails on a clean `app-redesign` too; it is a date bomb, not a regression here.

Browser-verified locally (dark first, then light, 1512px): header geometry, the blur ramp, the footer, the Total USDS and DAI 1Y and All ranges, and the staking reward rate All range.

Left for the preview deploy:
- Item 7 end-to-end against a real wallet with a Morpho position — the fix is root-caused on-chain and unit-tested, but the withdraw flow itself has not been run.
- Mobile layout of the footer — `resize_window` doesn't actually change the viewport locally.
- The BA Labs surfaces under real CORS. `info-sky.blockanalitica.com` only allows `app.sky.money`/`sky.money`, so every BA Labs chart is empty in local dev; the ranges above were verified through an in-page fetch stub reproducing the endpoints' real shape, page sizes and date spans.


inclinedmind (commented):
Reviewed the full sweep — the big removals (external-link interstitial, FooterLinks) check out clean with no dangling references, and the pagination walker, vault limits, and chart parsing are all well-tested. One ask, then some non-blocking notes.

**Safari ≤17 loses the header blur** — `globals.css` (`.progressive-blur-bar`): the mask is `-webkit-`-prefixed but `backdrop-filter` isn't, and unprefixed support only shipped in Safari 18. The old Tailwind `backdrop-blur-[7px]` emitted both, so this is a regression for older Safari: transparent bar, content colliding with the nav text. Add `-webkit-backdrop-filter: blur(20px);` — and worth one eyeball in an old Safari, since blur-inside-mask is a historically quirky WebKit path. If it renders badly there, an `@supports not (backdrop-filter: blur(1px))` solid-background fallback beats a glitchy blur.

**Non-blocking:**

- `VaultWidget` `isOverDepositCap`: while `assetBalance` is still undefined (initial load, network switch) with an amount already set, `maxDepositInput` resolves to `0n` and a false "over deposit cap" error shows — including on Morpho, which has no cap. The old code required `maxDeposit !== undefined` before firing. Gating on the balance being loaded would restore that.
- `useVaultTransactionForm`: `maxWithdrawInput ?? 0n` collapses the "unknown while liquidity loads" state back to zero, so the product-page modal still briefly flashes Balance: 0 for Morpho — the widget holds back instead, so the two surfaces this PR unifies still differ here. Separately, when liquidity-constrained, set-max fills the clamped amount but `max=true` routes the engine to redeem the *full* share balance — presumably caught by simulation, but worth a look.
- `useParseTokenChartData` 'All' range: the per-day loop calls `findMostRecentChange` (a linear scan) each iteration — O(n²), millions of comparisons for a 2019-start series, inside a main-thread `useMemo`. Fine at today's sizes; an index-tracking walk makes it O(n) whenever it starts to hurt.
- `PageFooter`: the legal-link filter is name-based (`/terms|privacy|risk/i`), so renaming a legal doc in `VITE_FOOTER_LINKS` silently drops it from the footer. Understandable given the env var carries no type field — just flagging the coupling.
- `fetchBaLabsPages`: the head-keeping `slice(0, wanted)` is only correct because the endpoints return rows newest-first — worth a one-line comment so a future ascending endpoint doesn't silently keep the oldest rows instead.

*Review written by Claude.*

PR #1818, opened 2026-08-06
feat(savings): supply USDC to Sky Savings on mainnet via the PSM

iceanddust:
## What

Adds **USDC as a supply origin for Sky Savings on mainnet**. The savings vault only takes USDS, so a USDC supply is the DAI path's shape with the USDS PSM wrapper standing in for the upgrade:

1. `approve` USDC → the PSM wrapper *(elided when the allowance covers it)*
2. `sellGem(user, gemAmt)` — the same call the Convert surface makes
3. `approve` USDS → sUSDS *(elided when covered)*
4. `deposit(usdsAmount, user, ref)` — the deposit every supply ends on

Bundled into one EIP-5792 call when the wallet supports it, otherwise up to four sequential transactions — exactly like `useBatchUpgradeAndSavingsSupply`.

**Withdraw is untouched:** mainnet withdraw stays USDS-only (static chip). The PSM leg here is supply-only, so a USDC exit is a withdraw to USDS followed by a Convert.

## Changes

| | |
|---|---|
| `hooks/savings/useBatchPsmSwapAndSavingsSupply.ts` | new engine, sibling of the DAI one |
| `useSavingsLaunch` | routes `isMainnetUsdc`; adds the 4-step labels |
| `SavingsOriginSelect` | `MAINNET_SUPPLY_ORIGINS` gains `'USDC'` |
| `useSavingsTransactionForm` | wad-correct preview read + the gate |
| `savings/hooks/useUsdcSupplyGate.ts` | new PSM availability gate (+ unit test) |
| `SavingsModalForm` | renders the blocked reason under the amount field |

One constant change carries the supply card too — its "Supply USDS / DAI / USDC" headline and its idle-balance sum both read `MAINNET_SUPPLY_ORIGINS`.

## Two things worth reviewing closely

**The 6 → 18 decimal seam.** USDC is 6-dec, both USDS legs are 18. The USDS approve *and* the deposit spend `math.convertUSDCtoWad(amount)`, and the orchestrator's `needsUsdsApproval` compares against that same wad — otherwise a small leftover allowance elides an approve step the engine still emits (covered by a test). The modal's `convertToShares` preview needed the same widening, or "You'll receive" showed ~0.

**Why a nonzero PSM fee is blocked, not just flagged.** `sellGem` returns `gemAmt × 1e12 × (WAD − tin)/WAD`. With a nonzero `tin` the swap under-delivers and the deposit can no longer be satisfied — and on the *sequential* path that leaves a swap that already landed. `useUsdcSupplyGate` therefore blocks on `tin > 0`, `live ≠ 1`, or a halted sell direction (reusing Convert's `getPsmDirectionHalted`), and holds the confirm until all three reads answer. Mainnet's `HALTED` reads as max-uint256, which that helper treats as "not halted" — now pinned by a test, since that sentinel is the branch production actually takes.

## Verification

Fork-verified end to end on a fresh Tenderly vnet, both wallet paths:

| | Sequential mock wallet | Batch mock wallet |
|---|---|---|
| Supplied | 1,000 USDC | 500.25 USDC |
| Outcome | 4 steps ✓, nonce +4 | "⚡ Bundled" badge, one step ✓ |

On-chain after both: USDC debited exactly (100,000 → 99,000 → 98,499.75), **USDS 0.000000 — nothing stranded**, sUSDS 1,355.70 matching the previewed 903.65 + 452.05. The position card flipped to "My position" live with no reload. Checked in dark mode; withdraw confirmed still USDS-only.

Tests: 9 calldata-parity / step-label tests (`useSavingsLaunch.usdc.test.tsx`), 11 gate-branch tests (`useUsdcSupplyGate.test.tsx`, mutation-checked), and 10 in the modal-form suite. Full suite green apart from the pre-existing `EarnFeaturedCards` maturity date bomb. Typecheck, lint and Prettier clean.

## Known residual

If the user rejects step 3 or 4 of the **sequential** path, the swap has already landed and they hold USDS in-wallet. Nothing is lost — they retry, or supply the USDS directly in one more click.

There is a second, rarer way to reach that same state, and it is **not** something the DAI path can do: `daiToUsds` has no fee, no `live` and no halt, so DAI's intermediate leg is unconditionally 1:1 and can only strand on a user rejection. The USDC leg's fee is governance-settable and the gate reads it off-chain with a 30s stale window (`refetchOnWindowFocus` on top). So if a spell set `tin` nonzero in the gap between that read and a sequential confirm, the first three calls land and the deposit no longer prepares. Batch is atomic and immune. Judged not worth pre-empting — `tin` has been 0 throughout and changing it is a governance action — but tightening `staleTime` on the wrapper reads at confirm time is the fix if we'd rather close it.

## Not in scope

- A `buyGem` mirror so savings can pay out in USDC (withdraw stays USDS-only by design).
- PSM debt-ceiling headroom. Convert treats mainnet USDC→USDS as unlimited liquidity and this follows that; an exhausted ceiling surfaces as a simulation failure rather than a friendly message.


inclinedmind (approved):
Approving — the money paths check out: wrapper address matches the chainlog `WRAPPER_USDS_LITE_PSM_USDC_A`, 6→18 decimal handling is exact everywhere (including the USDS-allowance-covers-the-6-dec-figure-but-not-the-wad edge, which has a regression test), approvals are exact-amount to the right spenders, the fee/halt gate covers `live`, the sell-direction bitmask and mainnet's max-uint `HALTED` sentinel, and L2 USDC still routes through psm3 with no conflict.

One non-blocking suggestion: the doc comment on `useBatchPsmSwapAndSavingsSupply` says `useSavingsLaunch` routes through `useUsdcSupplyGate`, but it doesn't — `useSavingsLaunch` arms the engine on `isMainnetUsdc` alone, and the gate only takes effect via the disabled confirm in `SavingsModalForm`. That leaves the zero-fee/halt invariant enforced in the UI layer only: a future caller of `launch()` (e.g. a portfolio quick-supply) would get an armed engine with no fee check, and with `tin > 0` the sequential path lands `sellGem` and then fails the deposit, stranding USDS. Suggest calling the gate inside `useSavingsLaunch` and folding `gate.ready && !gate.blockedReason` into the `enabled` it passes to the engine — the wrapper reads are already in the query cache from the form layer, so it's free. At minimum, correct the comment.

PR #1820, opened 2026-08-06
APP-484: hide geo-restricted products from the Portfolio page

inclinedmind:
Geo-restricted users could deposit into restricted modules (e.g. Sky Savings in the US) from the Portfolio page: the position cards' Supply buttons open deposit modals in place, skipping both the Earn listing partition and the product route guards.

Two layers of fix, matching the production precedent (APP-155) of hiding restricted modules outright — design confirmed: restricted products are hidden on Portfolio and surfaced only in the Earn module's "unavailable" section:

- **Hide restricted products from the Portfolio page.** A new `useGeoVisibleRows` hook (reusing the Earn page's `partitionByGeoAvailability`) filters the marketplace rows feeding the earnings card, position carousel, idle-table rate badges, and the disconnected marketplace section. The savings promos (allocate banner, TVL callout, idle-tab rate/projection stats) also drop when the savings module is restricted. The `/earn` page is untouched — it keeps its designed "Products unavailable in the US" section.
- **Gate the in-place supply modals in the resolver** (defense in depth): a position whose geo module is disabled resolves to no in-place handler, so the caller falls back to the product route, where the existing guard enforces the restriction.

**Testing / simulating a restricted region** (non-production builds): the geo override params apply to all of this — `?geo_mode=restricted` for the full restricted preset, `?geo_module_savings=false` for a single module, `?geo_country=US` to pin the country label. E2E fixtures can use the existing `mock-geo-config` helper.

Fixes APP-484.


apps/webapp/src/modules/portfolio/hooks/usePortfolioSupplyActions.ts
```diff
@@ -73,6 +80,9 @@ export function usePortfolioSupplyActions(): (position: SuppliedPosition) => (()
 
   return useCallback(
     (position: SuppliedPosition) => {
+      const geoModuleId = geoModuleForIntent(position.intent);
+      if (geoModuleId && !isModuleEnabled(geoModuleId)) return undefined;
```
iceanddust, line 87: **The three new gates disagree on what "geo config still loading" means.**

`useGeoVisibleRows` treats loading as *available* (returns rows unchanged), and so does `savingsAvailable = isGeoLoading || isModuleEnabled('savings')` in `ConnectedPortfolio`. This resolver has no such guard — and the provider answers `isModuleEnabled` → `false` for **every** module while `isLoading` (`GeoConfigProvider.tsx:36-40`).

So in the window between marketplace-loaded and geo-loaded, *every* Supply button — for unrestricted users too — resolves to `undefined` and navigates to the product page instead of opening the in-place modal. The cards are rendered in that window precisely because `useGeoVisibleRows` doesn't filter while loading.

The window is small and the fallback is safe, so this isn't a correctness bug so much as a policy that drifted from the two hooks it sits beside. Suggest making it explicit:

```ts
const { isModuleEnabled, isLoading: isGeoLoading } = useGeoConfig();
// …
if (!isGeoLoading && geoModuleId && !isModuleEnabled(geoModuleId)) return undefined;
```

plus `isGeoLoading` in the `useCallback` deps.

Worth noting the suite can't see this either way: neither this file's mock nor `PortfolioPositionsSection.test.tsx`'s includes `isLoading`, so the loading branch is untested.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
```diff
@@ -62,10 +70,13 @@ export function ConnectedPortfolio() {
...
   // and only after both data sources settle so it never flashes the wrong state.
+  // Visible totals only: a hidden restricted position mustn't suppress the
+  // callout or claim the Supplied tab for a portfolio that renders as empty.
+  const depositedUsd = visibleRows.reduce((acc, row) => acc + (row.position?.totalUsd ?? 0), 0);
   const idleTotalUsd = balances.reduce((acc, balance) => acc + balance.amountUsd, 0);
-  const callout = isLoading || balancesLoading ? 'none' : portfolioCallout(totalDepositedUsd, idleTotalUsd);
+  const callout = isLoading || balancesLoading ? 'none' : portfolioCallout(depositedUsd, idleTotalUsd);
 
-  const tab = userTab ?? (!isLoading && totalDepositedUsd <= SIGNIFICANT_BALANCE_USD ? 'idle' : 'supplied');
+  const tab = userTab ?? (!isLoading && depositedUsd <= SIGNIFICANT_BALANCE_USD ? 'idle' : 'supplied');
```
iceanddust: **Geo is now a third data source these gates need to wait on.**

The comment just above says the callout settles *"only after both data sources settle so it never flashes the wrong state"* — but `depositedUsd` now depends on geo too, and only `isLoading`/`balancesLoading` are in the gate.

For a US user holding sUSDS the page settles twice:
1. marketplace resolves → `visibleRows` still contains the savings row (loading is permissive) → `depositedUsd` is large → `tab` = `'supplied'`, `callout` = `'none'`
2. geo resolves → the row drops → the tab flips to `'idle'` under the user and the allocate banner pops in

Adding `isGeoLoading` to both gates makes it settle once:

```ts
const callout = isLoading || balancesLoading || isGeoLoading ? 'none' : portfolioCallout(depositedUsd, idleTotalUsd);
const tab = userTab ?? (!isLoading && !isGeoLoading && depositedUsd <= SIGNIFICANT_BALANCE_USD ? 'idle' : 'supplied');
```

`savingsAvailable` on line 45 has the same shape but degrades gracefully — it starts permissive and only ever hides, so it never flashes the wrong way.

iceanddust (commented):
Two findings from a review pass, both about how the three new gates treat the *loading* state of the geo config. Everything else I looked at (the reuse of `partitionByGeoAvailability`/`geoModuleForIntent`, the resolver as defense-in-depth, the new hook's test coverage) reads well — these are the two I'd want resolved before merge, and both are a one-line `isGeoLoading` addition.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
inclinedmind: Fixed in 39237cfc0 — added `isGeoLoading` to both the callout gate and the tab default as suggested, and updated the comment above them (three data sources now). Nice catch on the double-settle.

apps/webapp/src/modules/portfolio/hooks/usePortfolioSupplyActions.ts
inclinedmind: The divergence is real but intentional — though you're right it deserved to be stated. The display hooks fail open while the config loads so the page doesn't blank for everyone; this resolver fails closed so the loading window can't be used to open a restricted modal in place (the exact bypass this PR closes). The navigation fallback degrades safely for unrestricted users, so adding `!isGeoLoading` here would trade a sub-second modal-vs-navigate papercut for re-opening that window.

Made it explicit in 521d87f30: a comment at the gate, plus a test pinning the disabled-while-loading default in GeoConfigProvider.test — that contract turned out to be unpinned at its source (the provider suite never exercised the loading state), which seemed like the more valuable place for it than mirroring the provider's behavior in this file's mocks.

PR #1822, opened 2026-08-07
Remove the Chainalysis oracle fallback from the address check

iceanddust:
Wallet screening now relies solely on the `{authUrl}/address/status` endpoint.

## What changed

- **Deleted** `hooks/authentication/chainalysisOracle.ts` — the Chainalysis sanctions oracle `isSanctioned(address)` ABI and its address map across 10 chains.
- **Deleted** `hooks/authentication/useChainalysisCheck.ts` — the wagmi `useReadContract` wrapper around that oracle.
- **`useRestrictedAddressCheck`** collapses to a single `useQuery` against `/address/status`. The `chainId` prop is gone (it only existed to pick an oracle deployment).
- **`ConnectedProvider`** no longer calls `useChainId()` — it was only there to feed that prop.

No dependencies were involved; the oracle was read directly as a contract, so there is nothing to drop from `package.json`.

## Behaviour change worth reviewing

Previously the oracle was queried only when the HTTP check errored, and `useRestrictedAddressCheck` reported an error only if *both* failed. Now an `/address/status` failure surfaces as an error directly.

Because `ConnectedContext`'s `isAllowed` includes `!authError`, an API outage now sends connected users to the unauthorized state instead of falling through to the on-chain check. That is the intended consequence of relying solely on the endpoint — fail closed — but it removes the redundancy that kept the app usable if the API went down.

Note that address screening also still happens server-side on the terms endpoint: `checkTermsWithRetry` treats a 403 from `${VITE_TERMS_ENDPOINT}/check` as a deliberate access denial for a restricted region *or* a sanctioned address. That path is untouched. The IP-based `/ip/status` VPN/region check is also untouched.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` — 0 errors (278 pre-existing warnings)
- `pnpm -F webapp test` — 2357/2358 pass. The one failure is `EarnFeaturedCards.test.tsx`, the known hardcoded 18 Jun 2026 maturity date bomb that fails on every branch.
- `pnpm exec prettier --check` on both touched files — clean


fenris85:
*A single failed background refetch locks out an already-verified user, with no recovery path**

Fail-closed as described covers "the API is down and we can't get a verdict". What this change actually ships is stricter than that: it also discards a verdict we already have.

`QueryClient` is created with no `defaultOptions` (`lib/queryClient.ts:9`), so react-query defaults apply to the auth query: `staleTime: 0` and `refetchOnWindowFocus: true`. That means a verified user (cached `addressAllowed: true`) who tabs away and back triggers a background refetch, and if that one request fails - a transient blip, not an outage - react-query keeps the cached data but also sets `error`. Since `isAllowed` ANDs `!authError` (`ConnectedContext.tsx:157`), the cached approval is ignored and the full-app block kicks in.

There's also no way back: nothing in the UI re-triggers the `['auth', address]` query (`retryTermsCheck` only re-runs the terms check), and the query has no `refetchInterval`. The block is sticky until an unprompted future focus refetch happens to succeed or the user figures out that a manual reload helps. Previously the oracle fallback swallowed exactly this class of failure, so the path was effectively unreachable.

This degrades normal operation, not just outage behaviour, so I'd fix it in this PR. Any of these would do:

- give the auth query a sensible `staleTime` and `retry` policy so a single focus-refetch failure doesn't surface as an error;
- don't let `error` override a cached `addressAllowed: true` (only fail closed when there's no verdict at all);
- or add a retry path analogous to `retryTermsCheck` so the state is at least recoverable from the UI.

inclinedmind (approved):

PR #1824, opened 2026-08-10
Cache the settled portfolio decision per address + first-visit app loader + conditional landing (APP-419, APP-295)

codewiththeflowxyz:
## Summary

Follow-up to #1759, addressing the review concern there: the optimistic simulate callout can flash and then vanish for users whose real outcome is `allocate` or `none`. This PR makes the settled outcome **sticky per address**, so returning wallets get the right view on first paint, adds the **first-visit App Loader** from the design comps for the one case where nothing is cached yet, and builds the **conditional landing** (APP-295) on top of the same cache: `/` forwards to the visitor's likely home, and a manual connect sorts the user onto the right one behind the loader's cover.

### Decision cache (per address, 30-day TTL)

- When the portfolio queries settle, the resolved callout outcome + default tab are stored in localStorage (`portfolioDecision:v1:<address>`). The write runs at the shell on **every clean settle, on any route** — first load, a position-changing tx's refetch, a wiped cache repopulating — so the stored decision and its TTL stay current wherever the user is, not just on Portfolio.
- On the next visit that decision renders **instantly and stays frozen for the whole view** — fresh data only rewrites the cache, so a changed outcome applies on the next navigation, never as a mid-view swap.
- First visit (no cache or expired TTL) keeps the optimistic simulate fallback from #1759 unchanged.
- Module availability always wins over a cached outcome: cached promos wait for the settled module config instead of riding the optimistic default, so the banner can't paint and then vanish when the config lands. A manual tab pick always wins over the cached default; account switches remount onto the new address's cache.
- A decision is **never persisted from a failed fetch**: errored sources settle as empty data, indistinguishable from an empty wallet, so every writer (shell settle, Portfolio page settle, connect-time sort) skips and retries on the next clean settle. The shell writer also waits out the connect-time frame where a just-connected address's queries read settled before they have started, while a switch back to an address whose queries answer from the warm react-query cache (no loading cycle at all) still writes.
- The allocate banner now chips its $/year projection while figures load instead of flashing `$0/year`.
- Every write is mirrored to a `portfolioDecision:v1:$last` pointer (same TTL) that the `/` redirect reads synchronously.

### App Loader (Figma "App Loader" 1875:6834)

- One-shot ~2.2s timeline: the Sky wedge logomark (exported verbatim from the comp as `IllustrationSkyLogomark`) spins and pops over the bare page background, then chrome and content fade in on the house quint-out.
- Plays only on a genuine first visit, with two scopes: the **entry-gate play is once per browser** (a persistent played flag — no address is known yet while wagmi restores the session), and the **manual-connect play is once per wallet, at most once per page load** — the connecting address's cached decision gates it, so a new address gets covered while its own data fetches and sorts (an expired 30-day TTL replays it, since the data is genuinely refetched). A second wallet connecting later in the same page load skips the cover and sort — its settled decision is still cached by the shell writer, so its next connect resolves instantly. `prefers-reduced-motion` skips both entirely.
- The cover never traps input: the overlay and the hidden regions are `pointer-events-none`, so a dialog opening during the cover (e.g. the terms modal) stays fully usable while nothing invisible is clickable. A **2.5s watchdog** force-reveals if the exit timeline never reports completion (crashed subtree, dropped animation, background tab).
- Surfaces that mount outside `Layout` (the toast stack, the cookie banner, the dev commit-SHA line) hide under the cover via a document flag raised before the cover's first paint — hidden with opacity rather than visibility, so they keep focus and stay announceable to screen readers. A **network toast fired while covered defers to the reveal** and then shows with its full lifetime, instead of running out silently under a long hold; a newer network change supersedes a queued one (only the latest state shows, never a stale "Switched to X"), and unmount cancels a pending deferral.
- **Ship/no-ship is still a team decision** — `APP_LOADER_ENABLED` in `AppLoader.tsx` is the one-line toggle.

### Conditional landing + connect-time sorting (APP-295)

- `/` resolves in `beforeLoad` from the `$last` pointer: outcome `none` (the wallet had a significant position) → `/portfolio`; anything else, or nothing cached → `/earn`. Search params are preserved. The network-switch fallback now names `/portfolio` explicitly since `/` no longer means Portfolio.
- `$last` holds a copy of the most recently settled wallet's decision — `{ outcome, tab, updatedAt, address }` — where `outcome` is the settled callout state (threshold: $10 USD, so dust never flips it):
  - `none` — the wallet already has a meaningful earn position (> $10 deposited); no onboarding callout is shown, and its home is `/portfolio`.
  - `allocate` — no earn position, but > $10 of idle stablecoins; the allocate banner shows, and its home is `/earn`.
  - `simulate` — no earn position and nothing idle; the Sky Savings simulate pitch shows, and its home is `/earn`.
- On a **manual first connect** the loader's cover doubles as the landing sorter: it holds in a `held` mode — the logomark spin loops at the comp's angular speed — while the portfolio queries settle behind it, then the exit pop reveals the sorted home.
- Guard rails on the held cover: **1.2s minimum hold** (the spin reads as intentional and outlives the balances query's enable-flip frame) and an **8s cap** that releases a wedged hold without sorting — the settled outcome still gets cached whenever it arrives.
- The settle gate is a new `isPositionsLoading` flag on `useEarnMarketplace`: the decision only reads `row.position`, and the full `isLoading` also waits on external rate/chart APIs (BA Labs, Morpho, Pendle) that can outlive the cap.
- The shell-level queries (`useEarnMarketplace` + `useStablecoinBalances` + `useGeoConfig`) double as the app-level prefetch — react-query dedupes by key, so Portfolio/Earn mount warm — and feed the shell's background cache writer above.

#### Covered flows

| # | Flow | Wallet state | Loader | Data fetched behind it | User lands on |
|---|------|--------------|--------|------------------------|---------------|
| 1 | Unknown visitor opens `/` (nothing cached) | disconnected | none | — | `/earn` (default home) |
| 2 | Returning visitor opens `/`, `$last` outcome `none` | disconnected at redirect time (auto-reconnect may still be in flight — the redirect reads the cache synchronously, before wagmi resolves) | none | — | `/portfolio` |
| 3 | Returning visitor opens `/`, `$last` outcome `allocate`/`simulate` | same as 2 | none | — | `/earn` |
| 4 | Page entry while wagmi restores a session, nothing cached anywhere (genuine first visit) | reconnecting → connected (automatic) | **timed** — fixed ~2.2s Figma one-shot | prefetch warms marketplace + balances + geo (timeline doesn't wait on it) | route they entered |
| 5 | Manual first connect from `/portfolio` or `/earn`, no cache for that wallet | disconnected → connected (user clicks Connect) | **held** — spin loops until positions settle (min 1.2s, cap 8s), then exit pop | earn positions, stablecoin balances, prices, geo config (positions-only gate) | sorted home: `/portfolio` if significant position, else `/earn`; outcome + `$last` cached |
| 6 | Manual connect, wallet already has a cached decision | disconnected → connected (user clicks Connect) | none | — (cache hit) | instant sort to the cached home when sitting on the other one; otherwise stays |
| 7 | Manual first connect from a deeper page (e.g. `/stake`) | disconnected → connected (user clicks Connect) | **held** cover plays (visual only) | same as 5; outcome cached | never redirected — a connect inside a product flow stays put |
| 8 | Silent auto-reconnect on refresh | reconnecting → connected (automatic) | no held cover (entry gate may play flow 4) | prefetch only | stays put — auto-reconnect never sorts |
| 9 | Connect while terms not yet accepted | connected, terms pending | nothing until acceptance — modal first; rejecting disconnects | — | stays put; accepted → continues as 5 |
| 10 | Queries wedged past the 8s cap | connected (mid flow 5) | cover releases with the exit pop, **without** sorting | outcome still cached when it settles cleanly (never from a failed fetch) | stays put |
| 11 | Loader already played in this browser (`appLoader:v1:played` gates only the **entry-gate** play; the manual-connect cover is gated **per wallet** by the decision cache) | a different address connects manually | **held** cover plays on the page load's first manual connect — a new address means a fresh fetch and sort, so it's covered like flow 5. A later connect in the same page load isn't re-covered (the shell writer still caches its settle) | same as 5, for the new address | sorted home (flow 5); once its decision is cached, later connects use the instant sort (flow 6) |
| 12 | `prefers-reduced-motion` | any | never any cover | prefetch still runs | no held sort; instant-hint sort (flow 6) still applies |
| 13 | A data source errors during the held cover | connected (mid flow 5) | cover releases with the exit pop, **without** sorting | nothing persisted — a failed fetch is never cached | stays put; the next connect retries fresh |

## Test coverage

- **Unit**: cache round-trip/TTL/versioning/corrupt payloads, `$last` mirror/takeover + expiry, graceful no-cache degradation when storage is unusable, banner chip state; `/` destinations — Earn for unknown, Portfolio on `$last` `none`, Earn on `allocate`, search preserved.
- **Component**: instant render from cache, no mid-view swap + rewrite, settled writes for all three outcomes, no write on partial settle, expiry fallback, geo override, tab precedence, address switch; loader gates (entry, mid-session, played, reduced motion, terms pending), one-shot no-replay, cover surviving a failed reconnect; held-cover sorting — empty wallet sorts `/portfolio` → `/earn`, funded wallet stays + caches, instant cached-hint sort (re-fires on a repeat connect to the same target), matching-surface no-op, auto-reconnect ignored, deep pages never navigate, cap releases without sorting, a new address is still covered in a browser that already played, a settle whose live address changed since the connect releases without writing or sorting; review-round pins — the watchdog force-reveal, a failed source releasing without caching or sorting (both writers), and cached promos waiting for the settled module config; shell background writer — rewrites the settled decision on any route without a connect, skips errored settles, waits for the address's fetch cycle before its first write, and still writes on a warm-cache switch-back; covered-toast deferral — a toast fired under the cover shows at reveal with its full lifetime, a newer change supersedes the queued one, and unmount cancels a pending deferral.
- **Integration** (`AppLoaderTermsFlow`): the real `TermsModalProvider` + `TermsModal` + `TermsDialog` + `useAppLoader` chain — connect opens the modal with the loader held down and the checkbox locked behind the scroll gate; signing POSTs to the terms endpoint and only acceptance releases the loader; rejecting disconnects. This lives at the component level because `VITE_SKIP_AUTH_CHECK` auto-accepts terms in every e2e environment, so the modal is unreachable from e2e by design.
- **E2E** (`app-loader.spec.ts`, added to the green track): an unknown visitor's `/` lands on Earn; first connect with an empty wallet plays the held cover live, sorts to `/earn`, and caches `simulate` + `$last`; a returning wallet (cached decision) skips the loader and gets its entry rewritten on settle; a new wallet in a browser that already played still gets the held cover and lands sorted.

## Manual testing

Run `pnpm dev:mock` in `apps/webapp` and open `/portfolio`. The localStorage keys involved: `appLoader:v1:played`, `portfolioDecision:v1:<address>` and `portfolioDecision:v1:$last`. Reset between scenarios with:

```js
localStorage.removeItem('appLoader:v1:played');
Object.keys(localStorage).filter(k => k.startsWith('portfolioDecision')).forEach(k => localStorage.removeItem(k));
```

1. **First-visit loader + sort** — reset keys, refresh, click *Connect Mock Wallet*: the logomark spins while the data loads, then pops out and reveals the sorted home — `/earn` for a wallet without a significant position, `/portfolio` with one; the keys appear, the decision reads `{"outcome":"simulate","tab":"idle"}` for the default mock wallet.
2. **Conditional `/`** — after step 1, open `/` directly: it forwards to the home matching the cached outcome (and to `/earn` in a fresh incognito window).
3. **Returning user** — navigate to /earn and back: callout + Idle tab render instantly, no skeleton pass, no loader.
4. **Instant sort** — disconnect, go to the *other* home surface, reconnect the same wallet: no cover, you're moved to the cached home immediately.
5. **Deep pages stay put** — reset keys, open `/stake`, connect: the loader may play but you stay on `/stake`.
6. **Frozen outcome** — edit the decision value's `"outcome"` to `"allocate"`, reload: the allocate banner renders immediately (chipped figure fills in) and must NOT swap mid-view; after settle the cache is rewritten back — reload again and the simulate callout returns.
7. **TTL expiry** — set `updatedAt` to `1000000000000`, reload: first-visit behavior (optimistic simulate) returns and the cache refreshes.
8. **One-shot per wallet** — disconnect/reconnect with keys present: the loader never replays for the same wallet (the instant sort of step 4 still works). With a real wallet, switching to an address that has no `portfolioDecision:v1:<address>` entry plays the cover again for that address — its data has to be fetched and sorted too.
9. **Entry gate** — with a real wallet (`pnpm dev` + MetaMask): clear both keys and refresh — covered from first paint while wagmi reconnects.
10. **Reduced motion** — DevTools → Rendering → emulate `prefers-reduced-motion: reduce`, reset keys, connect: no loader, content renders directly.

Note: the terms modal cannot be reached manually with `VITE_SKIP_AUTH_CHECK=true` (local default) — that ordering is covered by the `AppLoaderTermsFlow` suite.

## Notes

- Stacked on #1759 — when that merges, this rebases onto `app-redesign`.
- The two items originally deferred to follow-up PRs are now included here: the app-shell prefetch (the shell-level queries) and the early-settle optimization (the positions-only loading gate).
- `tenderlyTestnetData.json` refresh included: the previous e2e vnets had expired (twice — they don't survive overnight).


apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,146 @@
...
+        animate={{ rotate: [0, -560, -640], scale: [1, 1, 1.6, 0] }}
+        transition={{
+          rotate: { duration: 1.6, times: [0, 0.75, 1], ease: [[0, 0, 0.3, 1], 'easeIn'] },
+          scale: {
+            duration: 1.6,
+            times: [0, 0.63, 0.75, 1],
+            ease: ['linear', [0, 0, 0, 1], [0.7, 0, 1, 1]]
+          }
+        }}
+        onAnimationComplete={onCoverEnd}
```
iceanddust: **medium** — the only exit from `cover` is `onAnimationComplete`; there is no fallback.

While `phase === 'cover'` the header, page content and mobile navbar all carry `opacity-0`. If `motion` never fires the callback — an error thrown inside the motion subtree (this overlay is not wrapped in an `ErrorBoundary`, unlike everything else in `Layout`), the animation being dropped, or a future `MotionConfig`/`skipAnimations` change — the app is permanently blank with no recovery path and no way for the user to get out.

A `setTimeout(endCover, ~2500)` armed when the cover starts bounds the worst case.

Related: because the timeline is rAF-driven (the per-segment cubic-bezier arrays force motion's JS loop rather than WAAPI), opening the app in a background tab leaves it fully invisible until the tab is focused.

iceanddust (commented):
Code review — 7 findings (4 medium, 3 low), split between the AppLoader cover overlay and the new portfolio decision cache.

Nothing blocking found in `portfolioDecisionCache.ts` itself (TTL, validation, corrupt-payload and unreadable-storage paths all look correct), in the `key={address}` remount in `PortfolioPage.tsx`, or in the `AllocateStablecoinsBanner` chip state.

apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,146 @@
...
+ * The covered stage: just the 64px logomark centered on the (already bare —
+ * everything else is opacity-0) page background. Timeline from the Figma
+ * motion spec, cut at the point the icon has shrunk away instead of looping:
+ * ~1.5 CCW turns easing out, a pop to 1.6x, then collapse to 0 by 1.6s.
+ */
+export function AppLoaderOverlay({ phase, onCoverEnd }: { phase: AppLoaderPhase; onCoverEnd: () => void }) {
+  if (phase !== 'cover') return null;
+  return (
+    <div
+      className="fixed inset-0 z-100 flex items-center justify-center"
```
iceanddust: **medium** — this transparent full-screen `z-100` overlay swallows clicks for any dialog open during the cover.

There's no `pointer-events-none`, and Radix dialogs in this repo are `z-50` (`components/ui/dialog.tsx:21,38`). The terms modal is mounted by `TermsModalProvider` *above* `Layout` in `pages/App.tsx`, so it is not hidden by the `opacity-0` content wrapper.

Concrete: the entry gate fires (`status === 'reconnecting'`, first visit); wagmi restores the connection ~200ms later and terms have not been accepted, so `TermsModal` opens — fully visible, with its own `backdrop-blur` scrim over the supposedly bare background — but every click lands on the loader overlay for the remaining ~1.4s.

The `AppLoaderTermsFlow` test only exercises the mid-session path, where the gate keeps `phase === 'off'`.

apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+const hasPlayed = () => {
+  try {
+    return localStorage.getItem(PLAYED_KEY) !== null;
+  } catch {
+    return true; // unreadable storage: assume played rather than risk repeats
+  }
+};
+
+const markPlayed = () => {
```
iceanddust, line 92: **low** — `markPlayed()` failures are undetectable, so the loader can replay on every load.

`hasPlayed()` fails safe (returns `true` when reads throw), but `markPlayed()` swallows write failures. On storage where `getItem` succeeds and `setItem` throws — Safari private browsing, a quota-full origin — the flag never persists, `hasPlayed()` stays `false`, and `writePortfolioDecision` fails for the same reason, so both gates keep passing: the app blanks out for 1.6s on *every* page load and every connect.

An in-memory `let played` fallback set alongside the write would at least bound it to one play per page load.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
```diff
@@ -78,18 +85,39 @@ export function ConnectedPortfolio() {
...
+    ? cachedDecision.outcome
+    : calloutLoading
+      ? 'none'
+      : portfolioCallout(depositedUsd, idleTotalUsd);
+  const optimisticSimulate = !cachedDecision && calloutLoading;
+
+  // Once every source settles, persist the computed decision for the next
+  // mount — this is the only thing fresh data changes mid-view, so it also
+  // refreshes an existing hint's outcome and TTL without touching the render.
+  useEffect(() => {
```
iceanddust, line 117: **medium** — a failed query is persisted as a settled decision and then frozen for the next visit.

`useStablecoinBalances` returns `isLoading: false` with `balances: []` when the balance query errors (`rawBalances ?? []`) *and* when only the price query errors (`parseFloat(priceFor(sym) || '0')` → every `amountUsd` is 0); `useEarnMarketplace` likewise yields `rows: []` on error. This effect only checks `calloutLoading`, so it cannot distinguish "settled empty" from "failed".

Concrete: a wallet with $500k idle USDC hits a transient prices outage, `{outcome:'simulate', tab:'idle'}` is written with a 30-day TTL; on the next visit that wrong decision is frozen for the *entire* view even though the data loads fine.

Skip the write when any source reports an error.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
```diff
@@ -162,12 +190,14 @@ export function ConnectedPortfolio() {
       {/* Banner and earnings card group: 56/48 below the header, and a tight
           16 between banner and card when a callout shows (1036:188968). */}
       <div className="mt-14 flex flex-col gap-4 md:mt-12">
-        {(calloutLoading || callout === 'simulate') && savingsAvailable && (
+        {(optimisticSimulate || callout === 'simulate') && savingsAvailable && (
           <SavingsTvlCallout tvlUsd={savingsTvlUsd} savingsRate={savingsRate} />
         )}
         {callout === 'allocate' && savingsAvailable && (
```
iceanddust: **medium** — a cached `allocate` renders a geo-restricted savings promo before the geo config lands.

`savingsAvailable = isGeoLoading || isModuleEnabled('savings')` is `true` while geo loads, and `callout` now comes straight from the cache instead of being forced to `'none'` during `calloutLoading`. So a user in a jurisdiction where `savings` is disabled, who has `{outcome:'allocate'}` cached, sees the full "Allocate stablecoins → $X/year" banner until `useGeoConfig` resolves, then it vanishes. Before this PR the allocate banner could never paint while geo was loading.

(The test at `ConnectedPortfolio.test.tsx:262` pins exactly this behavior for `simulate`, but only with `savingsEnabled: true`.)

Gating the cached outcome on `!isGeoLoading && isModuleEnabled('savings')`, or persisting a `savingsAvailable` flag alongside the decision, would close it.

apps/webapp/src/globals.css
```diff
@@ -314,6 +314,14 @@
   --animate-digit-roll-in: digit-roll-in 200ms cubic-bezier(0, 0, 0.58, 1);
   --animate-digit-roll-out: digit-roll-out 200ms cubic-bezier(0, 0, 0.58, 1) forwards;
 
+  /* First-visit AppLoader reveal (Figma 1875:6834): after the logomark
+     timeline, the page content fades in rising 30px and the chrome (navbars)
+     just fades, both on quint-out. Durations are the Figma timeline segments
+     (1.6-1.9s and 1.6-2.23s of the 3s loop). `both` holds the hidden state
+     through the frame between class swap and animation start. */
+  --animate-app-loader-content-reveal: app-loader-content-reveal 300ms var(--ease-out-quint) both;
```
iceanddust, line 322: **low** — `both` fill leaves a permanent non-`none` transform on `.page-transition`.

After the reveal the wrapper keeps `transform: translateY(0)` applied for the rest of the session (phase never leaves `reveal`, and `Layout` does not remount across routes). A non-`none` transform makes that element a stacking context, a containing block for `position: fixed` descendants, and a backdrop root — the exact hazard called out and avoided for the navbar wrapper at `Layout.tsx:120-121`.

Concrete: for a user who saw the loader, glass surfaces inside the page sample a different backdrop root than for everyone else (this repo has already been bitten by nested backdrop roots), and any non-portalled fixed child positions against the wrapper instead of the viewport.

Dropping the class once the animation ends (a terminal `done` phase) keeps the pre-animation hold without persisting the transform.

apps/webapp/src/modules/layout/components/Layout.tsx
```diff
@@ -106,8 +117,13 @@ export function Layout({
...
         <ErrorBoundary>
-          <MobileNavbar />
+          {/* Opacity-only wrapper: it never gets a transform, so it can't
+              become the containing block of the navbar's fixed pill. */}
+          <div className={appLoaderRevealClasses(loaderPhase, 'chrome')}>
+            <MobileNavbar />
+          </div>
         </ErrorBoundary>
+        <AppLoaderOverlay phase={loaderPhase} onCoverEnd={endCover} />
         <Banner />
```
iceanddust: **low** — `Banner` (and `CookieConsentBanner`) stay visible over the "bare page background" during the cover.

Only the header, `.page-transition` content and the mobile navbar get the reveal classes. `<Banner />` here, and `CookieConsentBanner` at `pages/App.tsx:82` (a sibling of `AppContent`, so outside `Layout` entirely), keep painting.

The cookie banner is a solid `bg-[#1a1a2e]` card that is shown *precisely* on a genuine first visit — i.e. exactly the scenario the loader targets — so the comp's clean logomark-on-background frame will have a consent card pinned bottom-right for the whole 1.6s.

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: Fixed in 2aefd813d: a 2.5s watchdog now forces the reveal whenever the cover's timeline never reports completion. Timed plays arm it from the start; held plays arm it once `released` — before that the 8s hold cap is the bound. This also bounds the background-tab case, since the watchdog is a plain timeout rather than rAF-driven. Test pin: "the watchdog reveals a cover whose timeline never completes".

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: Fixed in 2aefd813d: the overlay is now `pointer-events-none`, and the hidden chrome/content regions go inert too (`pointer-events-none opacity-0`) — so a dialog opening during the cover (the terms modal included) is fully usable, while nothing invisible becomes clickable through the transparent overlay.

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: The in-memory flag turns out to be redundant: within a page load the phase state machine already makes the play one-shot — once `phase` leaves `off` it never returns, pinned by the "is one-shot per page load" test — and across page loads no in-memory value survives anyway. On storage where writes throw, wagmi can't persist its connection either, so `status === 'reconnecting'` never fires the entry gate at all. The residual is the held cover on each manual connect in e.g. private browsing, which matches intent: no cached decision means the fetch genuinely needs covering.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
codewiththeflowxyz: Fixed in 2aefd813d: `useEarnMarketplace` now exposes `isPositionsError` (every source feeding `row.position`, prices included) and `useStablecoinBalances` exposes `isError` (balances + prices), and this persist effect skips while either is set. The AppLoader sort had the same bug — on error it now releases the cover unsorted and uncached, so the next connect gets a fresh try. Both paths have test pins.

apps/webapp/src/globals.css
codewiththeflowxyz: Already fixed on the branch in 32bcde01f (pushed after this review was cut): the keyframe's `to` frame now ends at `transform: none`, which `both` fill holds as identity — no stacking context, containing block, or backdrop root persists after the reveal. The comment in globals.css pins the reasoning.

apps/webapp/src/modules/layout/components/Layout.tsx
codewiththeflowxyz: `Banner` is fixed in 2aefd813d — it joins the chrome reveal, so it can't paint over the cover's bare frame. The CookieConsentBanner scenario can't occur, though: it auto-shows only after a flat 3.5s delay (`delayComplete`), and the entry cover is gone at ~1.6s, so the first-visit frame stays clean. We're deliberately not wiring it into the reveal — changing when that banner is presented is out of scope for this PR.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
codewiththeflowxyz: Fixed in b2d860078: the cached fast-path no longer rides the optimistic availability default — cached promos now wait for the settled module config (`!isGeoLoading && isModuleEnabled('savings')`), so the banner can't paint and then vanish when the config lands. The first-visit optimistic path is unchanged, and the test that pinned the old paint-through behavior now pins the wait.

apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,375 @@
...
+  // resolves the landing instantly instead; the navigation itself is a side
+  // effect and runs in the effect below.
+  if (pendingConnect && isConnectedAndAcceptedTerms) {
+    setPendingConnect(null);
+    const surface = sortableSurface(pathname);
+    const hint = readPortfolioDecision(pendingConnect);
+    if (hint) {
+      const target = landingFor(hint.outcome);
+      if (surface && target !== surface) setInstantTarget(target);
+    } else if (phase === 'off' && APP_LOADER_ENABLED && !prefersReducedMotion()) {
```
iceanddust, line 229: **medium — cover/sort only arm on the first connect per page load**

This arm requires `phase === 'off'`, but `phase` never returns to `off` (it goes `cover` → `reveal` and stays). `sortDoneRef`, `released` and `holdElapsed` are likewise never reset.

So only the *first* connect in a page load can be covered or sorted:

> first visit → connect wallet A (cover plays, sorts, `sortDoneRef = true`) → disconnect → connect uncached wallet B in the same session ⇒ no cover, no `SortJob`, no landing sort at all.

Flow 11 in the PR description claims a new address is always covered; the tests only exercise this from a fresh mount, so the in-session case is untested.

apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,375 @@
...
+  // in an effect) so the cover is up on the same frame. A cached decision
+  // resolves the landing instantly instead; the navigation itself is a side
+  // effect and runs in the effect below.
+  if (pendingConnect && isConnectedAndAcceptedTerms) {
+    setPendingConnect(null);
+    const surface = sortableSurface(pathname);
+    const hint = readPortfolioDecision(pendingConnect);
+    if (hint) {
+      const target = landingFor(hint.outcome);
+      if (surface && target !== surface) setInstantTarget(target);
```
iceanddust: **medium — `instantTarget` is state used as an event**

`setInstantTarget(target)` with the value already in state is a no-op, so the effect below (line 236) never re-runs.

> connect wallet A (cached `none`) from `/earn` → redirected to `/portfolio`; user walks back to `/earn`, disconnects, reconnects A → `setInstantTarget('/portfolio')` sets the same value, no re-render of that dep, no navigation.

The instant sort silently stops working for repeat connects to the same home.

iceanddust (commented):
Automated code review (`/code-review`) of the full PR diff — 28 files. 10 findings, posted inline below, roughly most-severe first.

The headline is a pattern rather than a single bug: `AppLoader` treats one-shot signals as sticky state, so the cover, the landing sort, and the instant redirect all only work on the **first** connect of a page load. The tests only mount fresh, so the in-session disconnect/reconnect path is uncovered.

These are review findings, not verified failures — worth checking each before acting.

apps/webapp/src/modules/app/hooks/useEnhancedNetworkToast.tsx
```diff
@@ -24,6 +24,28 @@ interface NetworkToastProps {
...
 }
 
+/**
+ * The app-loader cover hides the toast stack off the `data-app-loader-cover`
+ * document flag (globals.css), and the connect-time auto-switch fires exactly
+ * when the held cover goes up — shown immediately, a toast could burn its
+ * whole 5–8s lifetime invisible and never be seen. Hold the show until the
+ * flag clears (the cover always ends: settle, hold cap, or watchdog).
+ */
+const showWhenUncovered = (show: () => void) => {
```
iceanddust: **medium — `showWhenUncovered` breaks the one-toast-at-a-time dedup and leaks observers**

The `clearTimeout(toastTimeoutRef.current)` guard downstream can only cancel a *pending timeout*. Once a toast has been handed to a `MutationObserver` the ref is nulled, so a second network change while the cover is up queues a second observer and **both** toasts fire at reveal — including a stale "Switched to X" for a network the user has already left.

The observer is also never disconnected on unmount, or if `data-app-loader-cover` never clears.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
```diff
@@ -78,18 +91,41 @@ export function ConnectedPortfolio() {
...
+  // callout stands in while loading — its copy and button are static, only the
+  // TVL figure is data (chipped until it lands) — so the card doesn't pop in
+  // and shove the earnings card down. That pins the `simulate` outcome
+  // exactly; `allocate` (taller banner) and `none` (slot collapses) still
+  // shift — the winner can't be known while loading.
   const depositedUsd = visibleRows.reduce((acc, row) => acc + (row.position?.totalUsd ?? 0), 0);
   const idleTotalUsd = balances.reduce((acc, balance) => acc + balance.amountUsd, 0);
   const calloutLoading = isLoading || balancesLoading || isGeoLoading;
-  const callout = calloutLoading ? 'none' : portfolioCallout(depositedUsd, idleTotalUsd);
+  const callout = cachedDecision
```
iceanddust, line 105: **medium — cached `callout` (and `tab`, line 125) is frozen for the whole mount**

Pinning to `cachedDecision.outcome` leaves no escape hatch for data that changes *in view*.

> a user with a cached `simulate` decision supplies $500 through the portfolio's in-place supply modal — the "Sky Savings" onboarding pitch stays on screen and the tab stays on `idle` even though the wallet now has a position.

Before this PR the recomputed `portfolioCallout` would have dropped the callout. Consider unfreezing when the settled outcome differs *and* the change came from a local tx, or at least on `depositedUsd` crossing the threshold.

apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,375 @@
...
+  // at mount, so this never swaps a view; it only makes the next mount and
+  // the next `/` landing current. Armed by an observed loading cycle for this
+  // address, or by the balances query having fetched at least once under this
+  // address's key (`isFetched`) — an account switched back to inside gcTime
+  // resolves from warm cache without ever re-asserting isLoading, and its tx
+  // refetches must still land here. Both arms are false across the
+  // enable-flip frame, which is the case the guard exists for.
+  useEffect(() => {
+    if (!address || decisionLoading || decisionError) return;
+    if (fetchSeenForRef.current !== address && !balancesFetched) return;
```
iceanddust, line 187: **medium-low — the anti-enable-flip arm only observes the *balances* query**

`balancesFetched` can be true from a warm `token-balances` cache on an account switch-back, while the earn-position queries for the same address (`useMultiChainSavingsBalances`, `useAllMorphoVaultsUserAssets`, `useAllPendleUserAssets`, `useStUsdsData`) are still in their enable-flip frame reporting `isLoading: false` with no data.

The writer then caches `simulate`/`idle` and overwrites `$last` for a wallet that actually holds a large position — sending them to `/earn` from `/` if the tab is closed before the corrective write lands.

apps/webapp/src/modules/app/components/AppLoader.tsx
```diff
@@ -0,0 +1,375 @@
...
+    if (!sort || sortDoneRef.current || !holdElapsed || decisionLoading) return;
+    sortDoneRef.current = true;
+    // A failed source settles as empty data, indistinguishable from an empty
+    // wallet — caching that would freeze a wrong outcome for 30 days, and
+    // sorting on it could land a funded wallet on the simulate pitch. On
+    // error the cover just releases, unsorted and uncached; the next connect
+    // gets a fresh try.
+    if (!decisionError) {
+      const outcome = portfolioCallout(depositedUsd, idleUsd);
+      writePortfolioDecision(sort.address, {
```
iceanddust, line 282: **low-medium — persists/sorts for `sort.address` but reads balances for the *live* address**

`sort.address` is captured at `onConnect` time; `depositedUsd`/`idleUsd` come from whatever address is currently connected. The terms-gated path makes the divergence reachable:

> connect A → terms modal opens → user switches to account B in the wallet → accepts → `readPortfolioDecision`/`writePortfolioDecision` both run against A while every balance figure belongs to B.

Key the job off the live `address`, or bail when `address !== sort.address`.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
```diff
@@ -78,18 +91,41 @@ export function ConnectedPortfolio() {
...
+      ? 'none'
+      : portfolioCallout(depositedUsd, idleTotalUsd);
+  const optimisticSimulate = !cachedDecision && calloutLoading;
+
+  // Once every source settles, persist the computed decision for the next
+  // mount — this is the only thing fresh data changes mid-view, so it also
+  // refreshes an existing hint's outcome and TTL without touching the render.
+  // Never off failed data: an errored source settles as empty/zero totals,
+  // and caching that would freeze a wrong outcome for the TTL.
+  useEffect(() => {
```
iceanddust, line 117: **low — this second writer has none of the enable-flip protection given to the shell writer**

`AppLoader.tsx:182` got `fetchSeenForRef` + an `isFetched` arm; this one has neither, yet it writes the same per-address key *and* the shared `$last` pointer.

If the assumption behind the shell guard holds, this one can persist a zero-total decision on mount and poison `$last`. Either share one guarded writer, or apply the same arm here.

apps/webapp/src/routes/_shell.index.tsx
```diff
@@ -1,12 +1,22 @@
...
 // Legacy ?widget= / pre-flip-path deep links are already rewritten in __root's
-// beforeLoad (which throws first), so only a bare "/" reaches here; forward it
-// to /portfolio, preserving any global search params (e.g. ?network=).
+// beforeLoad (which throws first), so only a bare "/" reaches here. Global
+// search params (e.g. ?network=) are preserved.
 export const Route = createFileRoute('/_shell/')({
   beforeLoad: ({ search }) => {
-    throw redirect({ to: ROUTES.PORTFOLIO, search, replace: true });
+    const last = readLastPortfolioDecision();
+    const to = last?.outcome === 'none' ? ROUTES.PORTFOLIO : ROUTES.EARN;
```
iceanddust, line 19: **low — `readLastPortfolioDecision()` is address-agnostic**

`/` routes off *whoever settled last in this browser*, not who is connected.

> funded wallet A and empty wallet B on the same machine: opening `/` while B is the active wallet still lands on `/portfolio` (empty) until the shell writer rewrites `$last`.

The record already stores `address` — worth comparing it once wagmi resolves, or at least documenting the miss.

apps/webapp/src/modules/layout/components/Layout.tsx
```diff
@@ -28,6 +30,11 @@ export function Layout({
   const { siteConfig } = useContext(ConfigContext);
   const { chain } = useConnection();
   const { isConnectedAndAcceptedTerms } = useConnectedContext();
+  // First-visit loader (APP-419): while it covers, the chrome and content
+  // wear opacity-0 and the logomark overlay plays; on reveal they run their
+  // one-shot entrances. `off` leaves every className exactly as it was. On a
+  // manual first connect the cover also sorts the landing (APP-295).
+  const { phase: loaderPhase, coverMode, released, endCover } = useAppLoader();
```
iceanddust, line 37: **low — `useAppLoader` pulls the whole Earn marketplace into the shell**

It mounts `useEarnMarketplace` (BA Labs, Morpho rates + per-vault market data, Morpho trailing rates, Spark, Pendle markets + per-market history, stUSDS + chart) plus `useStablecoinBalances`, so every one of those external requests and contract batches now fires on `/stake`, `/convert`, and every widget route — not just Earn/Portfolio.

It's described as an intentional prefetch, but only `row.position` is consumed here. A positions-only hook would get the same gate at a fraction of the cost, and would keep BA Labs/Pendle failures off the critical path of unrelated pages.

apps/webapp/src/globals.css
```diff
@@ -377,6 +385,29 @@
       height: 0;
     }
   }
+  @keyframes app-loader-content-reveal {
```
iceanddust, line 388: **low — the comment reasons about the end state, but the `from` frame has a transform**

The `to` note explains why `transform: none` matters, and that's right. But `from` is `translateY(30px)`, so for the 300ms of the reveal the content wrapper *does* have a transform and *is* the containing block for `position: fixed` descendants.

Anything fixed rendered inside the page content during that window (a full-screen takeover or modal opened right as the cover lifts) is offset by the header for those frames.

iceanddust (approved):

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: Right — the cover and sort arm only on the page load's first manual connect, and we're keeping that: re-arming would mean resetting `released`/`holdElapsed`/`sortDoneRef` for an in-session wallet swap whose only loss is the cover visual and one redirect, while the shell background writer still caches the second wallet's settled decision — so its next connect resolves through the instant path. The description's flow 11 did overclaim this; it now reads "once per wallet, at most once per page load", with the shell-writer caveat spelled out.

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: Fixed in ff26b11de: the target is now held as an object (`{ to }`), so a repeat connect that resolves to the same landing gets a fresh identity and the navigation effect re-runs. Pinned by "the instant sort re-fires on a repeat connect to the same target" and verified in-browser: two consecutive /earn → /portfolio redirects for the same wallet, no cover either time.

apps/webapp/src/modules/app/hooks/useEnhancedNetworkToast.tsx
codewiththeflowxyz: Fixed in ff26b11de: the hook now tracks the deferred show in a ref — a newer network change disconnects and replaces a show still parked behind the cover, so only the latest state toasts at reveal, and unmount disconnects any pending observer and clears the timeout. Four unit pins (immediate show, deferral, supersede, unmount cancel), plus reproduced live: two chain switches under the cover produced exactly one toast at reveal, showing the final transition only.

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
codewiththeflowxyz: Keeping the freeze: no mid-view swap is the deliberate contract of this cache (the same reason the sort never yanks the user post-reveal), and the settle writer refreshes the record in the background, so the staleness is bounded to the current view — the next mount renders the corrected outcome. Unfreezing on local txs would reintroduce exactly the in-place content swap the design rejects; if this turns out to bite real users we can revisit with a scoped rule (e.g. only when a tx submitted from this view flips the outcome).

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: Acknowledged, keeping as-is: hitting it needs a warm balances cache alongside gc'd position caches and the tab closing inside the sub-second gap before the corrective settle write lands — any longer-lived session self-corrects, and the next page load's settle rewrites the record regardless. Threading `isFetched` through all six position sources didn't seem worth that residual; the loading-cycle arm still guards the flip frame in the common path.

apps/webapp/src/modules/app/components/AppLoader.tsx
codewiththeflowxyz: Fixed in ff26b11de: the settle effect bails when the live address no longer matches the job's (`address !== sort.address`) — it releases the cover without writing or sorting, so one wallet's totals can never be cached under another wallet's key. Pinned by "a settle whose live address changed since the connect neither writes nor sorts".

apps/webapp/src/modules/portfolio/components/ConnectedPortfolio.tsx
codewiththeflowxyz: Keeping this writer as-is for this PR: it only runs for a connected wallet on the Portfolio page, where the queries are either already observed loading (its settle gate skips `calloutLoading`) or warm from the shell, and the guarded shell writer — always mounted — re-asserts the correct record within the same settle cycle, while readers only consult the cache at mount. Agreed that one shared, guarded writer is the cleaner endgame; noted as a follow-up consolidation rather than growing this diff.

apps/webapp/src/routes/_shell.index.tsx
codewiththeflowxyz: By design: `/` redirects synchronously in `beforeLoad`, before wagmi has restored a connection, so there is no address to compare against without reintroducing the loading pass this landing exists to avoid. The record stores `address` precisely so a future consumer can compare once resolution exists at that point, and the shell writer rewrites `$last` on every settle, which keeps multi-wallet drift short-lived. The code comment documents the whoever-settled-last semantics.

apps/webapp/src/modules/layout/components/Layout.tsx
codewiththeflowxyz: Intentional — this is the approved app-level prefetch: the queries are shared by key with the Portfolio/Earn pages, so mounting them in the shell warms exactly what those pages request, and the decision gate reads only the positions-scoped flags, so external rate/chart failures never sit on the loader's critical path. A positions-only hook would trim the shell's request surface and is noted as a follow-up optimization; it duplicates query wiring we'd rather not fork inside this PR.

apps/webapp/src/globals.css
codewiththeflowxyz: True — for the 300ms of the content reveal the wrapper is a containing block, so a fixed surface opened inside the content during exactly that window is offset until the animation ends and `transform: none` clears it. Reaching it requires a takeover to open within those 300ms — input is inert until the reveal starts, so the window is effectively the animation itself. Accepting as a known, bounded limitation rather than restructuring the reveal.

PR #1826, opened 2026-08-11
APP-448: use the edited amount when retrying after a step-2 rejection and Back

inclinedmind:
## What

In sequential (bundling off) multi-step flows, a paused run's frozen call snapshot could be replayed after the user regained the editable inputs and changed them — signing a different amount than the form shows (APP-448). Two triggers: rejecting step 2 then pressing Back in the editable modals, and (for page-hosted engines like convert) rejecting step 2, closing the modal, and editing on the page.

## How

A wallet rejection pauses the run with the frozen snapshot still authoritative, so the next confirm resumed it and ignored the rebuilt calls. `TransactionContext` now bumps a `RevalidateRunContext` epoch at every editable-surface boundary — Back to the entry screen, close, session-replacing launch — and `useSequentialTransactionFlow` re-validates a paused run on the bump, dropping it once the live calls diverge from the frozen snapshot so the next confirm freezes whatever the form then shows.

An unchanged reconfirm still resumes mid-sequence — the paused run's index is the only memory that a mined value-moving first call (PSM `sellGem`, `upgrade`) already ran, since those are unconditionally rebuilt. Background rebuilds without a boundary crossing (a refetching quote) never reset, and `execute()` stamps the epoch so a reconfirm consumes the signal. The provider is app-wide on purpose: the divergence guard makes bumps safe for bystander engines.

## Testing

- Six unit tests in the sequential-flow harness: edit within the allowance, edit above it (rebuilds approve+supply), a quote-drift rebuild without a boundary keeping the resume, an unchanged Back resuming without re-running a mined `sellGem`, the reconfirm consuming the signal, and a bump clearing a run stranded by an on-chain revert.
- Re-enabled all four `sequential-tx.spec.ts` tests. The two stale-state canaries from #1729 cover the modal path end-to-end; the two happy-path tests had been parked for UI drift. The spec is rewritten for the current UI: bundling is seeded off via the persisted user setting (the switch moved to the nav menu), and the flow drives the three-screen entry → Review → Confirm modal with its inline failure + header-back recovery. Position checks assert deltas, never absolutes.


inclinedmind:
@coderabbitai review

inclinedmind:
@coderabbitai review

codewiththeflowxyz:
Reviewed the revalidation-epoch mechanism in depth. The core design (epoch bump at editable-surface boundaries + divergence check before resume) is sound for the tested paths, but the assumption **"calls diverged after a bump ⇒ the user edited"** breaks wherever calls drift mechanically, and two paths back to an editable surface never bump. Three findings are blockers.

## Blockers

**1. Unconsumed bump + mechanical call drift re-dispatches a mined value-moving call** — `useSequentialTransactionFlow.tsx:244`
A bump that finds the run resumable is never consumed (`runEpochRef` is only re-stamped in `execute()`), so the signal stays armed for the whole Back→reconfirm window. Any mechanical drift in that window — quote refetch (pendle 60s / curve 15s), or the allowance-consumption rebuild — is misread as a user edit and resets the run.
Concrete: sequential USDC→savings (`useBatchPsmSwapAndSavingsSupply`); `sellGem` mines, deposit rejected, user clicks Back **without editing**. The mined `sellGem` consumed the USDC allowance, so the rebuilt live calls reintroduce `approveUsdc` and match neither the frozen sequence nor its tail → `reset()` → next Confirm restarts at call 0 and **re-runs the mined `sellGem`, converting the user's USDC twice**. Unit tests cover churn-without-Back and Back-without-churn, but not Back+churn.

**2. Stranded runs reset even when calls are byte-identical** — `useSequentialTransactionFlow.tsx:248`
The keep-guard is gated on `isExecuting &&`, so a run stranded by an on-chain revert (`isExecuting=false, currentIndex>0`) is cleared by any bump unconditionally — but "Try again" does resume such runs, so "nothing to resume" doesn't hold. `[sellGem, supply]`: `sellGem` mines, `supply` reverts, user Backs to look, edits nothing, confirms → `sellGem` re-dispatched. This is also a regression: pre-PR, the confirm at `currentIndex=1` dispatched only the failed step. Fix: apply the `sameSequence` comparison to the `!isExecuting` branch too.

**3. Minimize/restore never bumps — APP-448 still reachable on page-hosted flows** — `TransactionContext.tsx:343`
Page-hosted engines (convert, pendle redeem, stake) keep their form editable under the minimized modal. Minimize during step 1 → step 2 rejected while minimized → edit the amount on the page → restore via the error toast → "Try again" → `handleRetry` takes the `isResume` branch and **signs the frozen old amount while the form shows the new one** — the exact bug this PR fixes, through a boundary that doesn't bump.

Findings 1, 2, and 5 share one root cause (the bump stays armed until the next `execute()`, and the reset path doesn't distinguish mechanical drift from edits nor exclude in-flight/stranded-but-resumable runs) — consuming/scoping the bump, guarding on a pending `mutationHash`, and dropping the `isExecuting &&` short-circuit likely clears all three.

## Should-fix

**4. Account switch never bumps the epoch** — `TransactionContext.tsx:220`
A paused run survives a wallet account switch; "Try again" has account B sign frozen calls whose args embed account A as receiver. The frozen-snapshot substrate predates this PR, but the PR builds exactly the mechanism that would catch it (post-switch live calls diverge) and omits the integration point.

**5. Reset can fire while a tx is unsigned in the wallet or mining** — `useSequentialTransactionFlow.tsx:251`
Close is allowed during INITIALIZED and the revalidation effect never checks for a pending mutation. Close (bump) + background quote drift → `reset()` mid-flight clears the mutation hash; if the user signs the still-open prompt, the tx mines unobserved (`onSuccess`/balance refetch never fire) and the engine at index 0 can re-dispatch on the next confirm. Hardening: skip the reset while a dispatched write is unresolved.

**6. e2e `positionAmount` baseline race** — `sequential-tx.spec.ts:50`
Both position cards render the supply CTA whenever balance data is `undefined` (no loading guard), and `connectAndVerify` only waits for the wallet chip — so `positionBefore` can read 0 before the balance query resolves. On a fork account with a pre-existing position, the re-enabled canaries fail or flake for reasons unrelated to APP-448.

## Minor

**7.** `useSequentialTransactionFlow.tsx:17` — `sameCall`'s "the abi is deliberately ignored, it only affects encoding" comment is a false generalization: the abi selects the overload and drives calldata encoding. No live trigger today, but worth correcting the comment (or comparing `encodeFunctionData` output).

**8.** `sequential-tx.spec.ts:34` — `seedBundlingOff` hardcodes `'user-settings'`, duplicating `USER_SETTINGS_KEY` (`lib/constants.ts:138`). If the key is renamed, the seed silently stops applying and all four tests exercise the batch path. A comment linking the two locations is the zero-cost fix.

**9.** `revalidateRunContext.ts:3` + bump sites — the same resume/drop rationale is restated in four multi-line comment blocks; suggest one-line invariants in source with the full matrix in the PR description.

Also verified and fine: the bystander-engine concern with the app-wide provider (no page-hosted engine has a value-moving non-final call, worst case is a redundant approve prompt), the `handleRetry` no-bump behavior (correct — retry must resume), the barrel export, and the re-enabled e2e specs' locators/seeding.

PR #1828, opened 2026-08-11
Restore the stUSDS max-withdraw resync in the redesigned modal (APP-507)

inclinedmind:
Restores a behavior lost in the widget→modal port, per APP-507.

The retired widget kept a Max-filled withdraw amount synced to the live max while idle; the modal froze it at click time even though the Curve quotes behind it refetch every 15s. This adopts the savings-form pattern instead of the widget's resync effect: the `max` flag (not the typed number) drives the redemption and is exempt from the amount gating, and the displayed value *derives* from the live max at render time — no resync state to freeze after an error, flicker the confirm on quote refetches, or rewrite the input during a transient quote reload (confirm is held while the Curve max quote is loading). The ≥2% price-impact acknowledgement is now keyed to the amount it was given for, so a drifting max lapses the checkbox and re-gates confirm.

Unit tests cover the derived display (native + Curve drift), gating exemption, loading hold, acknowledgement lapse, and the typed-value/supply paths (10 cases). Typecheck, eslint, prettier clean; stusds suite (35) passes.

### DIFF
--- a/apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
+++ b/apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
@@ -102,34 +102,15 @@ export function useStUsdsTransactionForm({
   // balance (no dust) and the Curve quote runs in max mode.
   const [max, setMax] = useState(false);
 
-  const amount = parseAmount(value);
-  const debouncedAmount = useDebounce(amount);
-  const debouncePending = debouncedAmount !== amount;
-
   const { data: stUsdsData } = useStUsdsData();
   const { data: capacityData } = useStUsdsCapacityData();
-
-  const providerSelection = useStUsdsProviderSelection({
-    amount: debouncedAmount,
-    referenceAmount: REFERENCE_AMOUNT,
-    direction: isSupply ? StUsdsDirection.SUPPLY : StUsdsDirection.WITHDRAW,
-    userStUsdsBalance: stUsdsData?.userStUsdsBalance,
-    isMax: max
-  });
-  const isCurveSelected = providerSelection.selectedProvider === StUsdsProviderType.CURVE;
 
   const {
     effectiveBalance: withdrawBalanceLimit,
     curveMaxWithdraw,
-    selectedProvider: withdrawSelectedProvider
+    selectedProvider: withdrawSelectedProvider,
+    isLoading: withdrawBalancesLoading
   } = useStUsdsWithdrawBalances();
-
-  // If Curve can absorb the supply there's no protocol limit on the input;
-  // otherwise the native module capacity caps it (widget behavior).
-  const isCurveAvailableForSupply = providerSelection.curveProvider?.state?.canDeposit ?? false;
-  const moduleMaxSupplyAmount = isCurveAvailableForSupply
-    ? undefined
-    : (providerSelection.nativeProvider?.state?.maxDeposit ?? capacityData?.remainingCapacityBuffered ?? 0n);
 
   // Native max withdraw is buffered against liquidity drift; the Curve max is
   // the pool quote for the full share balance.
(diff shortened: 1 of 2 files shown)


inclinedmind:
@coderabbitai review

inclinedmind:
@coderabbitai review

apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
```diff
@@ -179,7 +203,13 @@ export function useStUsdsTransactionForm({
     }
   };
 
-  const amountReady = isConnected && !isZero && !insufficient && !blocked && !debouncePending;
+  // Max readiness needs a real max: zero (nothing to withdraw) or a still-loading
+  // Curve max quote (whose transient fallback is the smaller native max) hold
+  // the confirm rather than gating on the derived number.
+  const amountReady =
```
iceanddust, line 214: The max branch drops `!isZero && !debouncePending`, so the CTA can go live while `debouncedAmount` is still `0n` — and every amount-derived surface then reads `0.00 USDS` for a full-balance redemption.

On a fresh Withdraw modal `value` is `''`, so `debouncedAmount` starts at `0n`. Clicking Max/100% makes `available > 0n` immediately, and `disabled` then only waits on `prepared` — the fresh `redeem` simulation, which doesn't depend on `amount` at all. That eth_call can easily return inside the 500 ms debounce (and returns instantly on a second Max within 30 s, since `useWriteContractFlow` caches it with `gcTime: 30000`).

In that window the review grid renders `Amount 0.00 / Receive 0.00` and `positionAfter` equals the current position. On its own that self-corrects within half a second, since the `updateModalContent` push stays live while `txStatus === IDLE`. It only becomes sticky if Confirm also lands inside the window: `useModalEntryBody` freezes on the first non-IDLE status, so the wallet screen and both toasts stay pinned at "Withdrawing 0.00 USDS" / "0.00 USDS withdrawn!" for a tx that redeems the whole share balance.

That needs Max → Review → Confirm inside 500 ms, so it's an unlikely sequence rather than a likely one — but it's cheap to close. Adding `&& debouncedAmount > 0n` to the max branch gates only the never-settled state and leaves the drift-lag exemption (and the "no confirm flicker on drift" test at `useStUsdsTransactionForm.test.tsx:164`) intact. The Curve route is already protected incidentally: `selectedQuote` is `undefined` at `amount === 0n`, so `curveWithdraw` is disabled and `prepared` is false.

apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
```diff
@@ -102,13 +102,37 @@ export function useStUsdsTransactionForm({
...
+
+  const available = isSupply ? (stUsdsData?.userUsdsBalance ?? 0n) : maxWithdrawAmount;
+
+  // Withdraw Max is flag-driven (savings-form pattern): the engine redeems the
+  // whole share balance / swaps the live quote, never the typed number, so the
+  // displayed amount derives from the live max instead of a click-time
+  // snapshot — the Curve quotes behind `available` refetch every 15s, and a
+  // pure derivation keeps tracking them with no state rewrite to freeze,
+  // invalidate, or misfire (the retired widget resynced with an effect).
+  const displayValue = max && !isSupply ? formatUnits(available, DECIMALS) : value;
```
iceanddust, line 131: The live derivation re-keys the amount-keyed native preview read on every drift, so on a Curve-backed max the Confirm button and the provider notice pulse into a disabled/loading state roughly every 15 s.

Chain: `available` for a Curve-selected withdraw is `curveMaxWithdraw`, whose `get_dy` carries `refetchInterval: 15000` (`useCurveQuote.ts:129`). Each refetch moves `available` → `displayValue` → `amount` → `debouncedAmount`, which re-keys `useStUsdsPreviewWithdraw(amount)` — a plain read with no `placeholderData` (`useStUsdsPreviewWithdraw.ts:18-25`). Its `data` drops to `undefined` and `isWithdrawPreviewLoading` goes true, so `providerSelection.isLoading` is true for one RPC round trip — and `disabled` in `StUsdsModalForm.tsx:124` includes it.

A user parked on the review screen sees Confirm grey out every ~15 s; a click landing in that gap is swallowed and has to be repeated, and `StUsdsProviderNotice` drops to its loading state at the same cadence. This is new here — the click-time snapshot kept the query key constant.

Worth flagging because it's the same failure mode the `amountReady` exemption above was added to prevent; `disabled` just has a second door into it. Either `placeholderData: keepPreviousData` on the preview read, or exempting `max` from the `providerSelection.isLoading` term, delivers the rest of that intent.

In passing: during that gap the native quote degrades to the `isValid: false` stub with `effectiveRate: 0n`, which sends `compareRates` down its `rateB === 0n` path (a hard 100% difference). Harmless while the notice is behind `isLoading`, but it's why the route would flip if the two ever got out of sync.

apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
inclinedmind: Good catch — fixed in 6bcb591. The max branch now also requires `debouncedAmount > 0n`, which gates exactly the never-settled window: on a fresh form, Confirm holds until the debounce settles off 0, and on a drift re-derivation the debounced value still carries the previous settled amount, so the drift-lag exemption (and the no-flicker-on-drift test) are unchanged. Added a test for the fresh-form window.

apps/webapp/src/modules/stusds/hooks/useStUsdsTransactionForm.tsx
inclinedmind: Fixed in 6bcb591 with `placeholderData: keepPreviousData` on `useStUsdsPreviewWithdraw` (the `useNetworkFee` idiom). Each drift re-key now hands back the previous preview while the fresh one loads, so `isLoading` never pulses and `disabled` holds steady on review — and the `effectiveRate: 0n` stub stays out of `compareRates` mid-refetch, closing the route-flip door you flagged in passing.

Went with the placeholder over exempting `max` from the `isLoading` term: it fixes the pulse at the source rather than adding a second exemption, and the exemption route would have traded the button pulse for a hero-placeholder flicker on review (the selected quote's `inputAmount` drops in the same window).

On typed edits the stale-pairing window this could theoretically open doesn't surface: the Curve `get_dx` read re-keys on the same amount change with no placeholder, so `providerSelection.isLoading` still covers the round trip and the UI shows a genuine loading state, same as before. Added a test pinning the placeholder config plus the zero-amount gating.

